PR #4, opened 2025-06-24
  Add minimal Dockerfile and docker-compose.yml for containerized deploy

topiaruss:
A minimal approach to a multiplatform docker build. 

Currently the build script targets `topiaruss` - my account. You will not be able to push there, so choose another destination. There are a couple of occurences of that string, BTW

The build worked for me on an M3 mac. It is not yet tested on amd64. 

My testing was limited to `haiku-rag serve` which indexed a few docs. 

You can take it from there. Happy to help further...

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,10 @@
+__pycache__/
+*.pyc
+*.pyo
+*.db
+.git
+.gitignore
+tests/
+database/
+documents/
+uv.lock 
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,7 @@ tests/data/
 
 # environment variables
 .env
+
+# dirs we might mount in docker-compose.yml
+documents/
+database/
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,28 @@
+FROM python:3.11-slim
+
+# Install uv using pip instead of the shell installer - to avoid platform issues
+RUN pip install uv
+
+# Set working directory
+WORKDIR /app
+
+# Copy project files
+COPY . .
+
+# Install the package
+RUN uv pip install --system .
+
+# Create directories for mounts
+RUN mkdir -p /inbound /database /ollama/models
+
+# Set environment variables
+ENV MONITOR_DIRECTORIES="/inbound"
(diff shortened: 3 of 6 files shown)


ggozad:
@topiaruss so very nice Russ!
I am gonna have a look and test it out asap!

topiaruss:
get a free account here:  https://hub.docker.com/repository/docker/ggozad
public FOSS images are free. You get (?)1 free private image.

ggozad:
There are some things not working I think:
When running the container on a silicon Mac you get 
```
OperationalError: /app/.venv/lib/python3.10/site-packages/sqlite_vec/vec0.so: 
wrong ELF class: ELFCLASS32
```

I believe this is related to https://github.com/asg017/sqlite-vec/issues/159 which is closed with https://github.com/asg017/sqlite-vec/pull/160
So supposedly it should be fixed upstream, but from what I gather there hasn't been a release of `sqlite-vec` since the merge....

topiaruss:
which command were you running to see the error?

Since there are many libs only exercised by certain commands, or with certain models, It would be *really* useful to have a test suite matrix of actual commands that are real integration tests on each model, to help trigger lurking arch issues. 

Would you like me to sketch something?

ggozad:
> which command were you running to see the error?
> 
> Since there are many libs only exercised by certain commands, or with certain models, It would be _really_ useful to have a test suite matrix of actual commands that are real integration tests on each model, to help trigger lurking arch issues.
> 
> Would you like me to sketch something?

I get it just running `haiku-rag`, it happens when we try to add the vector extension to sqlite. So it's kind of fundamental but I think it's on `sqlite-vec` end as I mentioned above.
Tests would be useful although I think the libs we use are pretty cross-compatible. Let's manage to figure out `sqlite-vec` on `aarch64` first perhaps.

topiaruss:
Will take a look.

ggozad:
This can be revisited now @topiaruss if you want. We have replaced `sqlite-vec` that did not compile properly with `lancedb`.

ggozad:
This was done in #100

PR #6, opened 2025-06-24
feat(tests): Add tests for app and cli modules

topiaruss:
Boosting your coverage a bit.
If you don't like the style, I will not be offended. ;)


ggozad:
Thank you so much Russ, fixed a couple of things that made pyright unhappy.

Issue #43, opened 2025-08-31
instruction to install sqlite with extensions


silentyak:
By default, the sqlite in macOS does not include the extensions.

```
export SQLITE_PREFIX=$(brew --prefix sqlite)
export CPPFLAGS="${CPPFLAGS} -DSQLITE_ENABLE_LOAD_EXTENSION=1"
export CFLAGS="${CFLAGS} -DSQLITE_ENABLE_LOAD_EXTENSION=1"
PYTHON_CONFIGURE_OPTS="--enable-loadable-sqlite-extensions" \
LDFLAGS="-L${SQLITE_PREFIX}/lib" \
CPPFLAGS="-I${SQLITE_PREFIX}/include" \
pyenv install -v YOUR_PYTHON_VERSION_HERE
```

PR #44, opened 2025-09-02
Replace sqlite & sqlite-vac with lancedb.

### LINKED ISSUE #41: sqlite-vec is, at least temporarily, dead
The https://github.com/asg017/sqlite-vec repo has not had any updates in 7 months despite over 20 PRs. Let's face it, at the rate the LLM world is moving that's basically the same thing as a decade in other tech areas.

I see a few forks of sqlite-vec but none of them look very promising.

I would however, like to point out that there is first party vector similarity search in DuckDB: https://duckdb.org/docs/stable/core_extensions/vss.html

I have not tried this, and do not claim it is a good move, but wanted to surface it in case you were not aware.

Thank you for your FOSS work!

ggozad:
This is a major change that completely replaces using sqlite in favor of [lancedb](https://lancedb.com)

* All existing functionality works the same way.
* We gain as a result of the transition a significant boost in accuracy: from 67% to 85% with the same basic model settings.

In addition:

* We deprecate and remove the Ollama reranker. Reranking using Ollama will only be introduced again if Ollama properly supports it.

Closes #41, #43


tpoindex:
Looks good, better accuracy is a plus.

On a side note, it doesn't look like lancedb supports in-memory db's.  Not a show stopper, however.

PR #178, opened 2025-12-04
Support for `import_document()` to cater for external pipelines

ggozad:
- **BREAKING: `create_document()` API**: Removed `chunks` parameter
  - `create_document()` now always processes content (converts, chunks, embeds)
  - Use new `import_document()` for pre-processed documents with custom chunks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,8 +24,8 @@
   - Requires terminal with image support (iTerm2, Kitty, etc.)
 - **New `import_document()` Method**: Import pre-processed documents with custom chunks
   - Use when document conversion, chunking, and embedding were done externally
-  - Accepts `content`, `chunks` (required), and optional `docling_document_json`/`docling_version`
-  - Validates that if docling parameters are provided, both must be present
+  - `chunks` required; `content` optional if `docling_document_json` is provided
+  - When `docling_document_json` is provided without `content`, content is extracted from the DoclingDocument
   - Validates docling JSON parses correctly if provided
 - **`update_document_fields()` DoclingDocument Support**: Added `docling_document_json` and `docling_version` parameters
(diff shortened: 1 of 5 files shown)


ggozad:
We also need to care about `update_document_fields()`.
Having it accept docling document as a parameter, gives us these cases:
  | Case | content | chunks | docling_document_json | What should happen                                      |
  |------|---------|--------|-----------------------|---------------------------------------------------------|
  | 1    | ✗       | ✗      | ✗                     | Only update metadata/title, no rechunk                  |
  | 2    | ✓       | ✗      | ✗                     | Convert content → DoclingDocument, rechunk              |
  | 3    | ✗       | ✓      | ✗                     | Use custom chunks as-is, no docling data                |
  | 4    | ✗       | ✗      | ✓                     | Extract content from docling, rechunk from it           |
  | 5    | ✓       | ✓      | ✗                     | Use custom chunks, ignore content for chunking          |
  | 6    | ✓       | ✗      | ✓                     | Conflict: which content? Error or use docling?          |
  | 7    | ✗       | ✓      | ✓                     | Use custom chunks + store docling JSON                  |
  | 8    | ✓       | ✓      | ✓                     | Conflict: Use custom chunks, but which content/docling? |

I think the way to handle each case is as follows:

  | Case | Behavior                                                             |
  |------|----------------------------------------------------------------------|
  | 1    | Only update metadata/title                                           |
  | 2    | Convert content → DoclingDocument, rechunk, update content           |
  | 3    | Use custom chunks, update content if provided via chunks             |
  | 4    | Extract content from DoclingDocument, rechunk using that docling     |
  | 5    | Use custom chunks, update content field                              |
  | 6    | Error: Providing both content and docling_document_json is ambiguous |
  | 7    | Use custom chunks + store docling JSON (like import_document)        |
  | 8    | Error: Same as 6                                                     |

PR #334, opened 2026-04-02
Build RAG docs database on release

ggozad:
- Adds a GitHub Actions workflow that builds a LanceDB database from docs/ on each release
- Uses `haiku-ingester run-batch`: ingests new and changed docs, and deletes docs that were removed from the tree
- Restores the previous release's database and queue, so only changed docs are re-processed and deletions are detected
- Uploads the database and queue as a release artifact (`haiku-rag-docs.tar.gz`)

### DIFF
--- a/.github/haiku.rag.yaml
+++ b/.github/haiku.rag.yaml
@@ -1,3 +1,6 @@
+storage:
+  vacuum_retention_seconds: 0
+
 embeddings:
   model:
     provider: ollama
--- a/.github/haiku.rag.yaml
+++ b/.github/haiku.rag.yaml
@@ -1,0 +1,8 @@
+embeddings:
+  model:
+    provider: ollama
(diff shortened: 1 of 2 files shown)


tseaver (commented):
I wonder:  could this PR also emit as an artifact a `haiku-rags-docs` skill, generated from the `haiku-rag create-skill` command and the ingested lancedb?

ggozad:
> I wonder: could this PR also emit as an artifact a `haiku-rags-docs` skill, generated from the `haiku-rag create-skill` command and the ingested lancedb?

It could. I haven't pushed further on this mostly because I am not sure in which shape to ship it. It could be an artifact that is used as part of a tutorial for example, but needs some docs work.
But yes, once the db is ready the skill could be generated by the template. I will revisit asap.

PR #374, opened 2026-05-17
feat: add '--search-type' option to CLI 'search'

tseaver:


ggozad:
Thanks for picking this up.

- `search_type` should be a `Literal["hybrid", "fts", "vector"]` (or `Enum`) end to end, defaulting to `"hybrid"` at the CLI. That removes both the `str | None` vs `str = "hybrid"` mismatch across layers and the silent fallback when an invalid value is passed.
- The image branch in `client.search` is always vector-only, either reject `--search-type` with `--image` or note it in the help text.
- Add a `--search-type` example to docs/cli.md.

ggozad:
Few more things:

* Widen the annotation to `SearchType | None` and pass `None` as the default:

```
search_type: SearchType | None = typer.Option(
    None,
    "--search-type",
    "-s",
    help="Type of search for text queries: vector, fts, or hybrid (default: hybrid)",
),
```

This is because `--image` queries now error out. The Typer default is "hybrid", so search_type is never None by the time it hits the app-layer check:

```
if query is None and search_type is not None:
    self.console.print("[red]Pass --search-type only for text queries[/red]")
```

* Since store sits below client, I would move `SearchType` into `store` (or inline in `chunk.py`) and have the client import from there.
* Small typo "onlly" in `client/search.py`

tseaver:
@ggozad I think I've addressed everything.

PR #391, opened 2026-06-01
perf: event-driven worker wakeup via asyncio.Condition

mcdonc:
## Summary
- Workers previously polled the queue with a fixed 1s sleep between claim attempts, adding ~500ms average latency to job pickup
- `JobRepo.job_available` (an `asyncio.Condition`) is now notified on every successful enqueue, waking idle workers immediately
- The poll interval remains as a timeout fallback for stop signals and breaker state changes

## Test plan
- [ ] Verify job pickup latency improvement under load

### DIFF
--- a/haiku_rag_slim/haiku/rag/ingester/queue/repository.py
+++ b/haiku_rag_slim/haiku/rag/ingester/queue/repository.py
@@ -66,6 +66,9 @@ class JobRepo:
         # JobRepo and SyncStateRepo share the same connection, callers must
         # pass the same lock instance so cross-repo calls also serialize.
         self._lock = lock or asyncio.Lock()
+        # Notified after a successful enqueue so workers can wake
+        # immediately instead of polling on a fixed sleep interval.
+        self.job_available = asyncio.Condition()
 
     async def enqueue(
         self,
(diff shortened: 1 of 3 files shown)


ggozad:
The wakeup idea is good. One regression though: this slows shutdown.

Idle workers now park on `job_available.wait(timeout=poll_idle_s)`, but `stop()` only sets `_stop` — it never notifies `job_available`. So an idle worker sleeps out the full interval before noticing `_stop`. The old `_sleep_or_stop` waited on `_stop` directly and woke instantly. Measured with the default `poll_idle_interval_s=1.0`:

```
main:  stop() with idle workers = 0.007s
#391:  stop() with idle workers = 0.802s
```

Fix is to wake the waiters when stopping:

```python
async def stop(self) -> None:
    self._stop.set()
    async with self._jobs.job_available:
        self._jobs.job_available.notify_all()
    ...
```

Also worth a test or two: that an idle worker picks up an enqueued job in ≪ `poll_idle_s` (the feature), and that `stop()` stays fast with idle workers (guards this regression). Right now only the existing suite runs, which wouldn't catch either.

PR #392, opened 2026-06-01
perf: add partial indexes for has_pending() and dashboard queries

mcdonc:
## Summary
- `has_pending()` scans `jobs WHERE source_id=? AND status IN ('queued','claimed')` on every poller sweep — full table scan as jobs grow
- `count_succeeded_since()` scans by `completed_at` for the dashboard's rolling-throughput display
- Adds two partial indexes: `idx_jobs_pending_by_source` and `idx_jobs_succeeded_completed`
- Both use `CREATE INDEX IF NOT EXISTS` so they're idempotent on existing databases

## Test plan
- [ ] Verify EXPLAIN QUERY PLAN uses the new indexes

### DIFF
--- a/haiku_rag_slim/haiku/rag/ingester/queue/schema.py
+++ b/haiku_rag_slim/haiku/rag/ingester/queue/schema.py
@@ -40,6 +40,18 @@ ON jobs(scheduled_at)
 WHERE status = 'queued'
 """
 
+JOBS_PENDING_BY_SOURCE_INDEX = """
+CREATE INDEX IF NOT EXISTS idx_jobs_pending_by_source
+ON jobs(source_id)
+WHERE status IN ('queued', 'claimed')
+"""
+
+JOBS_SUCCEEDED_COMPLETED_INDEX = """
+CREATE INDEX IF NOT EXISTS idx_jobs_succeeded_completed
+ON jobs(completed_at)
+WHERE status = 'succeeded'
+"""
+
 SYNC_STATE_DDL = """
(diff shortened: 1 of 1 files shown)


ggozad:
`idx_jobs_succeeded_completed` is a clear win, `count_succeeded_since()` was doing a full `SCAN`, and this turns it into a covering index search. 

`idx_jobs_pending_by_source` is redundant though. `has_pending()` filters on `source_id` + `status IN ('queued','claimed')`, and the existing `uq_jobs_live` partial index is already `(source_id, uri) WHERE status IN ('queued','claimed')`. `source_id` is its leading column, so the planner already uses it. 

EXPLAIN on a throwaway DB:

```
-- before any new index
has_pending:           SEARCH jobs USING INDEX uq_jobs_live (source_id=?)   -- already indexed
count_succeeded_since: SCAN jobs                                            -- full scan

-- after both new indexes
has_pending:           SEARCH jobs USING INDEX idx_jobs_pending_by_source (source_id=?)
count_succeeded_since: SEARCH jobs USING COVERING INDEX idx_jobs_succeeded_completed (completed_at>?)
```

So `has_pending()` was never doing a full table scan — the "full table scan as jobs grow" in the description isn't accurate. Adding the second index just shifts it to a marginally narrower index at the cost of write amplification on every insert/claim/complete, which is the hot path this series is trying to speed up.

Shall we drop `idx_jobs_pending_by_source` and keep just `idx_jobs_succeeded_completed`?

PR #393, opened 2026-06-01
perf: batch sync_state writes during poller sweeps

mcdonc:
## Summary
- Each discovered file previously triggered a separate `sync.upsert()` with its own lock acquire + SQLite commit (fsync)
- On a sweep finding 1,000 files this meant 1,000 individual commits at ~1-5ms each
- Adds `SyncStateRepo.batch_upsert()` that flushes all rows in a single transaction
- `_sweep_once` now collects sync_state rows into a list and calls `batch_upsert()` once at the end

## Test plan
- [x] 4 new tests for `batch_upsert()` (insert, update, preserve-on-None, empty noop)
- [ ] Verify reduced commit count under load with large source directories

### DIFF
--- a/haiku_rag_slim/haiku/rag/ingester/pollers/base.py
+++ b/haiku_rag_slim/haiku/rag/ingester/pollers/base.py
@@ -129,11 +129,13 @@ class BasePoller:
                     SourceEventKind.DELETE: 0,
                     SourceEventKind.UNCHANGED: 0,
                 }
+                sync_batch: list[tuple[str, str, str | None, str | None, bool]] = []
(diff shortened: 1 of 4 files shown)


ggozad:
One nit: the positional `tuple[str, str, str | None, str | None, bool]` recurs in the annotations, the three `append`s, and the unpack in `batch_upsert`. For the sake of future clairty can we use a small `NamedTuple`?

```python
class SyncRow(NamedTuple):
    source_id: str
    uri: str
    revision: str | None
    content_hash: str | None
    ingested: bool
```

PR #395, opened 2026-06-01
perf: reuse httpx.AsyncClient in HTTP and WebDAV sources

mcdonc:
## Summary
- HTTPSource and WebDAVSource previously created a new `httpx.AsyncClient` for every `head()`, `fetch()`, and `discover()` call
- Now create the client once in `__init__` and reuse it across all requests — enables TCP connection reuse and avoids repeated TLS handshakes
- Both sources gain an `aclose()` method; `PollerManager.stop()` calls it on shutdown to cleanly close connection pools

## Test plan
- Existing tests for both sources exercise head/fetch/discover through mock transports — behavior is unchanged

### DIFF
--- a/haiku_rag_slim/haiku/rag/ingester/app.py
+++ b/haiku_rag_slim/haiku/rag/ingester/app.py
@@ -181,6 +181,7 @@ class IngesterApp:
                     await asyncio.gather(api_task, return_exceptions=True)
                 await self._pollers.stop()
                 await self._stop_pool()
+                await self._pollers.close_sources()
 
     async def run_batch(self) -> BatchReport:
         """Run one discover() sweep across every configured source, drain the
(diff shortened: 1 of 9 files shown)


ggozad:
lifecycle has a shutdown-order bug.

Workers share the same `Source` instances as the pollers (`app.py`, `sources=self._pollers.sources`) and use them for `fetch()`. 

In `serve()` the teardown closes the clients *before* the workers stop:

```python
finally:
    await self._pollers.stop()   # aclose()s every source's httpx client
    await self._stop_pool()      # workers keep draining in-flight jobs for up to shutdown_grace_s
```

So during the grace window a worker finishing (or claiming) an UPSERT calls `fetch()` on an already-closed client → `RuntimeError: Cannot send a request, as the client has been closed`. The clients need to be closed only after the pool is fully stopped, e.g. keep `aclose` out of `PollerManager.stop()` and close sources after `_stop_pool()`, or just stop the pool first.

Also `run_batch()` never closes them at all — its `finally` only calls `_stop_pool()`, never `_pollers.stop()`, so the clients created in `__init__` leak each run. Whatever the fix, make sure both paths close the clients after workers are done.

Minor: `if hasattr(source, "aclose")` — worth promoting `aclose()` to the `Source` protocol (no-op default for FS/S3) so it's explicit and type-checked.

PR #396, opened 2026-06-01
perf: stagger periodic poll sweeps to avoid thundering herd

mcdonc:
## Summary
- All pollers sharing the same `poll_interval_s` previously woke and swept at exactly the same moment after startup
- With 10+ sources this causes coordinated spikes in listing traffic (S3 LIST, WebDAV PROPFIND, HTTP HEAD) every interval
- Add a random initial delay of 0–25% of the poll interval after the first sweep, applied to both `PeriodicPoller` and `FSPoller`'s sweep loop
- The initial sweep on startup still runs immediately; only subsequent sweeps are staggered

## Test plan
- The `run()` methods are marked `pragma: no cover` (event-loop glue); sweep behavior is tested via `_sweep_once()` directly

### DIFF
--- a/haiku_rag_slim/haiku/rag/ingester/pollers/base.py
+++ b/haiku_rag_slim/haiku/rag/ingester/pollers/base.py
@@ -1,5 +1,6 @@
 import asyncio
 import logging
+import random
 from datetime import UTC, datetime
 
 from haiku.rag.config import SourceConfig
@@ -14,6 +15,8 @@ from haiku.rag.ingester.sources.base import (
 from haiku.rag.telemetry import get_context, logfire
 
 logger = logging.getLogger(__name__)
+
+_STAGGER_FRACTION = 0.25
 
 
 def _enqueue_extra() -> dict | None:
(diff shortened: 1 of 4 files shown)


ggozad:
Logic looks right. The jitter block is duplicated verbatim in `fs.py` and `periodic.py` though since both subclass `BasePoller` (which already holds `config` and `_stop`), it'd be cleaner as one helper there:

```python
async def _stagger_start(self) -> bool:
    """Sleep a random fraction of the interval so pollers sharing an
    interval don't sweep in lockstep. Returns True if stop fired."""
    jitter = random.uniform(0, self.config.poll_interval_s * _STAGGER_FRACTION)
    try:
        await asyncio.wait_for(self._stop.wait(), timeout=jitter)
        return True
    except TimeoutError:
        return False
```

That also gives a seam to unit-test the stagger (monkeypatch `random`, set `_stop`) without the `# pragma: no cover` event-loop glue. And the `0.25` should be a named constant.

PR #399, opened 2026-06-01
fix: handle file deleted during discover() stat() call

mcdonc:
## Summary
- A file deleted between `os.walk()` and `path.stat()` in `FSSource.discover()` raises `FileNotFoundError`
- This propagated uncaught, failing the entire sweep and tripping the circuit breaker
- Now catches `FileNotFoundError` around the `stat()` call and skips the file — the next sweep or watchfiles will emit the DELETE event

## Test plan
- [x] 1 new test: deletes a file mid-iteration and verifies discover() completes without error

### DIFF
--- a/haiku_rag_slim/haiku/rag/ingester/sources/fs.py
+++ b/haiku_rag_slim/haiku/rag/ingester/sources/fs.py
@@ -139,7 +139,12 @@ class FSSource:
             if not self.filter.include_file(str(path)):
                 continue
             uri = path.as_uri()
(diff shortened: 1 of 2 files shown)


ggozad:
Fix is correct. But: the test passes without it (I reverted the `try/except` to check) as a file deleted before the loop reaches it is already caught by `path.is_file()`, so the new `except` never runs. To actually cover the race, monkeypatch `Path.stat` to raise `FileNotFoundError` once while `is_file()` returns `True`.

PR #405, opened 2026-06-01
fix: run_batch hangs forever when all workers die

mcdonc:
## Summary
- The drain loop in `run_batch()` polls `counts_by_status()` waiting for queued/claimed to reach zero
- If all worker tasks crash, claimed jobs stay claimed forever and the loop never exits — the `run-batch` CLI command hangs
- Now checks `live_workers` during the drain loop; if claimed jobs exist but no workers are alive, logs an error and breaks out
- Stranded jobs will be reaped by the reaper on the next start

## Test plan
- [x] 1 new test: kills all workers while a job is claimed, verifies `run_batch` completes within timeout instead of hanging

### DIFF
--- a/haiku_rag_slim/haiku/rag/ingester/app.py
+++ b/haiku_rag_slim/haiku/rag/ingester/app.py
@@ -204,12 +204,13 @@ class IngesterApp:
                     counts = await self._jobs.counts_by_status()
                     if not counts.get("queued") and not counts.get("claimed"):
                         break
-                    if counts.get("claimed") and self._pool.live_workers == 0:
+                    if self._pool.live_workers == 0:
+                        outstanding = counts.get("queued", 0) + counts.get("claimed", 0)
                         logger.error(
-                            "All workers have died with %d claimed job(s) — "
-                            "aborting batch; stranded jobs will be reaped on "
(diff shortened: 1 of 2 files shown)


ggozad:
Two issues before this lands:

**Condition misses the `queued` case.** If all workers die while jobs are still `queued` (not yet claimed), we get `queued > 0`, `claimed == 0`, `live_workers == 0` and it hangs anyway. Want:
```python
if (counts.get("queued") or counts.get("claimed")) and self._pool.live_workers == 0:
```

**Test doesn't hit the new branch.** `create_document_from_source` is an `AsyncMock`, so `side_effect = lambda: stall.wait()` returns
an un-awaited coroutine instead of blocking, the job finishes instantly and is never durably claimed. And even if it blocked, `_process` catches `CancelledError` and releases the claim back to `queued`, so `task.cancel()` can't leave anything `claimed`. The killer never sees a claimed job, `run_batch` exits the normal way, and `assert report is not None` passes regardless (the abort log never fires).

The real trigger is a hard worker crash. Test should kill workers that way (e.g. patch `run_job` to raise a plain `Exception`), then assert it returns within a timeout and logs the abort.

PR #419, opened 2026-06-04
Update to monty 0.0.19

ggozad:
Move the analysis sandbox to `pydantic-monty` 0.0.19.

The previous `>=0.0.18` pin already resolved to 0.0.19 on a fresh lock, and 0.0.19 removed `MontyRepl`. The pin and the code have to move together.

### Migration

- Sandboxed code runs in a subprocess worker pool. `MontyRepl` becomes `AsyncMonty()`, then `session = pool.checkout(limits=...)`, driven by `session.feed_run(...)`. The `external_functions=` argument becomes `external_lookup=`.
- `Sandbox.close()` is now async. It returns the worker to the pool and shuts the pool down. `AnalysisCapability._close` awaits it.
- The document VFS does not change. `OSAccess`, `CallbackFile`, and `MemoryFile` work as before.
- `pydantic-monty-runtime` arrives as a transitive dependency, so packaging does not change.
- A worker crash returns a failed `SandboxResult` instead of raising out of `execute()`. The handler catches `pydantic_monty.MontyError`.
- Text that the sandbox writes to stderr now lands in `SandboxResult.stdout`. The `stderr` field carries the exception.

### open() and with blocks

Document files accept `open()` and `with` blocks. File objects support `.read()`, `.readline()`, and `.readlines()`. A write raises `PermissionError`.

A file object is still not iterable. `for line in f` raises `TypeError`. See pydantic/monty#490, which is still open. The analysis instructions state this limit in three places, and each one gives the alternative: `.readlines()` or `.read().split("\n")`. A new test pins the limit. That test fails when monty adds iteration support, which is the signal to relax the instructions.

The `collections` module is still unavailable.

### Timeout enforcement

`analysis.code_timeout` never limited one execution. Two changes make it do that.

- The VFS bridge suspends the worker for the length of a read, and Monty checks its budget between interpreter steps. It cannot check while a read is in flight. Code that reads in a loop overran a 60s budget by minutes. The sandbox now checks the deadline before each read. It raises from inside the callback, which answers the worker's suspension and keeps the session usable.
- Monty spends `max_duration_secs` across the session rather than per call, and the sandbox reuses the session so that variables persist. The session budget is now `code_timeout * max_executions`. At the old per-call value the first slow call starved every later one.

`feed_run` stays outside `asyncio.wait_for`. Cancelling during pure compute is clean. Cancelling while a read waits for an answer wedges the session with a protocol `RuntimeError` that escapes `execute()`. A call that computes without reading stays bounded by the session budget alone.


ggozad:
Continued in #521. The branch was renamed to `feat/monty-0.0.19`, which closed this PR.

Issue #436, opened 2026-06-11
pdf attachment mime misrouting


bd-mkt:
# PDF embedded-attachment MIME misrouting (`.joboptions` → docling PDF backend)

## Summary

When ingesting a PDF, haiku.rag extracts its embedded files
(`/EmbeddedFiles`) and ingests each as a child document. The file
extension that decides how each attachment is parsed was derived from the
child's **synthetic URI**, not from the attachment's own filename. Because
that URI puts the attachment name in a URL *fragment*, the extension
fallback fell through to the **parent PDF's `.pdf` suffix**. As a result,
plain-ASCII attachments such as Adobe's `Press Quality.joboptions` were
written to a `*.pdf` temp file and fed to docling's PDF backend, which
fails to parse them.

## Root cause

The child URI is built with the attachment name in the fragment
(`client/documents.py`, `_reconcile_pdf_attachments`):

```python
name = att.get_name()                       # e.g. "Press Quality.joboptions"
child_uri = f"{parent_doc.uri}#attachment={quote(name, safe='')}"
content_type = mimetypes.guess_type(name)[0] or "application/octet-stream"
```

`.joboptions` is unknown to `mimetypes`, so `content_type` becomes
`application/octet-stream`. The extension was then derived in
`_ingest_fetch_result` via `get_extension_from_content_type_or_url`
(`client/processing.py`):

```python
for ct, ext in content_type_map.items():    # octet-stream: no match
    if ct in content_type:
        return ext
parsed_url = urlparse(url)                   # url = ...#attachment=...
path = Path(parsed_url.path)                 # fragment dropped -> parent path
if path.suffix:
    return path.suffix.lower()               # -> ".pdf"  (the PARENT's!)
return ".html"
```

Worked example:

| step | value |
|------|-------|
| attachment name | `Press Quality.joboptions` |
| `mimetypes.guess_type(name)` | `(None, None)` |
| `content_type` | `application/octet-stream` |
| `child_uri` | `file:///docs/brochure.pdf#attachment=Press%20Quality.joboptions` |
| `urlparse(child_uri).path` | `/docs/brochure.pdf` (fragment dropped) |
| derived suffix | **`.pdf`** |

The `.pdf` extension passed the `supported_extensions` guard, the ASCII
bytes were written to a `*.pdf` temp file, and docling routed them to
`DoclingParseDocumentBackend`, raising
`ValueError("Failed to parse file: ...")`. That `ValueError` was **not**
[...]

PR #437, opened 2026-06-11
fix for issue with parsing pdf attachment such as *.joboptions files

bd-mkt:
Fix for issue #436


ggozad:
Thanks @bd-mkt, added a couple of small edits and made tests pass in CI.

PR #448, opened 2026-06-18
improve concurrency management for pdf attachments

bd-mkt:
# fix: dispatch `_extract_pdf_attachments` via `asyncio.to_thread`

## Problem

`_reconcile_pdf_attachments` called `_extract_pdf_attachments` synchronously
from the async pipeline:

```python
# before
new_attachments = _extract_pdf_attachments(parent_body, parent_doc.uri, depth=depth)
```

`_extract_pdf_attachments` is a synchronous function that holds `PDFIUM_LOCK`
(a `threading.Lock`) for its entire body — opening the PDF with
`pdfium.PdfDocument(bytes)`, iterating every embedded attachment, and
extracting each attachment's bytes via `att.get_data()`. Running this on the
event-loop thread blocks the loop for the full duration of pdfium I/O.

During that block:
- All other concurrent worker coroutines are suspended, regardless of where
  they are in the pipeline.
- Any `asyncio.to_thread` results waiting in the thread pool (e.g. completed
  docling conversions) cannot be delivered back to their awaiters.
- The `asyncio.sleep(0.1)` polling loop in `run_batch` cannot fire.

In `run-batch` mode the full corpus is queued in one sweep, so all four workers
start processing simultaneously. When several workers reach PDFs with embedded
attachments at roughly the same time, the event loop freezes repeatedly —
once per document with attachments, serially — producing the thousand-second
store-step delays observed in practice.

## Fix

Dispatch the call through `asyncio.to_thread`, matching the pattern already
used by `convert_pdf_with_splitting` for `_next` and `it.close`:

```python
# after
new_attachments = await asyncio.to_thread(
    _extract_pdf_attachments, parent_body, parent_doc.uri, depth=depth
)
```

`_extract_pdf_attachments` is self-contained (no async internals) and
thread-safe under `PDFIUM_LOCK`, so the only requirement for correctness in a
thread is that the lock is acquired before any pdfium call — which it already
is throughout the function body. The event loop is now free to service other
coroutines while the thread waits for and holds the lock.

## Test

`test_extract_pdf_attachments_called_off_event_loop_thread` in
`tests/test_pdf_attachments.py` installs a spy around
`_extract_pdf_attachments` that records `threading.current_thread()` at call
time, then asserts the recorded thread is not `threading.main_thread()`. This
directly tests the dispatch mechanism: a regression back to a synchronous call
will fail the assertion with a clear message, independent of document size or
timing.

## Files changed

- `haiku_rag_slim/haiku/rag/client/documents.py` — `import asyncio`; wrap
  `_extract_pdf_attachments` call in `await asyncio.to_thread(...)`
- `tests/test_pdf_attachments.py` — new threading-identity test

### DIFF
--- a/haiku_rag_slim/haiku/rag/client/documents.py
+++ b/haiku_rag_slim/haiku/rag/client/documents.py
@@ -1,3 +1,4 @@
+import asyncio
 import hashlib
 import json
 import logging
@@ -543,7 +544,9 @@ async def _reconcile_pdf_attachments(
     if (parent_doc.metadata or {}).get("content_type") != "application/pdf":
         return
 
-    new_attachments = _extract_pdf_attachments(parent_body, parent_doc.uri, depth=depth)
+    new_attachments = await asyncio.to_thread(
+        _extract_pdf_attachments, parent_body, parent_doc.uri, depth=depth
+    )
     if new_attachments is None:
         return
 
--- a/tests/test_pdf_attachments.py
+++ b/tests/test_pdf_attachments.py
@@ -1,10 +1,12 @@
 import io
+import threading
 
 import pypdfium2 as pdfium
 
 from haiku.rag.client import HaikuRAG
 from haiku.rag.client.documents import (
     MAX_ATTACHMENT_DEPTH,
+    _extract_pdf_attachments,
     _reconcile_pdf_attachments,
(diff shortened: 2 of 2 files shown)


ggozad:
Thanks @bd-mkt, looks good!

PR #464, opened 2026-06-23
adjust zstd handling to avoid possible core dumps with concurrency

bd-mkt:
Change zstd pattern to better handle concurrent threads doing compression. Also enabled multithreaded compression

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Changelog
 ## [Unreleased]
 
+### Fixed
+
+- zstd compression uses a fresh `zstandard` compressor/decompressor per call instead of shared module-level singletons, fixing process segfaults when ingester workers compress documents concurrently (Python < 3.14).
+
 ## [0.61.0] - 2026-06-23
 
 ### Added
--- a/haiku_rag_slim/haiku/rag/store/compression.py
+++ b/haiku_rag_slim/haiku/rag/store/compression.py
@@ -17,7 +17,7 @@ except ImportError:
     # _prepare_document_from_docling), so construct a fresh instance per call
     # rather than sharing a module-level singleton.
     def _zstd_compress(data: bytes) -> bytes:
-        return ZstdCompressor(threads=-1).compress(data)
+        return ZstdCompressor().compress(data)
 
     def _zstd_decompress(data: bytes) -> bytes:
         content_size = get_frame_parameters(data).content_size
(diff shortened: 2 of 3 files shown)


ggozad (commented):
Thanks for this, the core fix is correct, onstructing a fresh instance per call is the right fix. 

I'd like to drop the `threads=-1` part, though, and keep `ZstdCompressor()`. It's a separate change from the thread-safety fix, and after measuring it it looks like it works against the huge-PDF memory pressure we are doing this for.

What I measured (zstandard 0.25.0, cext backend):

- **It's a no-op for typical blobs.** Multi-threaded compression only engages
  once the input exceeds zstd's job-size threshold. At 0.5/2/8 MB, `threads=-1`
  ran single-threaded anyway (cpu/wall ≈ 1.0). It only activates on the very
  large page-raster blobs — i.e. exactly the image-heavy documents where the
  downside below lands hardest.
- **It raises peak memory.** On a 24 MB blob (measured as process maxRSS in
  isolated subprocesses):

  | config | peak RSS delta |
  |--------|----------------|
  | `ZstdCompressor()` | +18 MB |
  | `ZstdCompressor(threads=-1)` | +61 MB |

  That's ~3× the compression memory, from zstd's per-worker window/job buffers
  (C-side, so `tracemalloc` doesn't show it) — and it's per concurrent ingestion
  worker. For a service that's getting OOM-killed on huge-PDF batches, adding
  tens of MB × `worker_count` of transient peak is the wrong direction.
- **Over-subscription.** It spins up ~ncpu compression threads per call
  (cpu/wall ≈ 2.8) on top of our N ingestion worker threads.
- **The only upside doesn't apply.** MT shaved wall time on the 24 MB blob
  (≈10 ms vs 26 ms), but ingestion already parallelizes across documents/workers,
  so single-call compress latency isn't the bottleneck. We'd be paying memory +
  CPU oversubscription for wall-time we don't need.

Sanity note since I was surprised by the differing output sizes: 
MT compresses in independent jobs and finds fewer cross-job matches, so on long-range-redundant input it can produce a larger frame; both decompress correctly. For our high-entropy base64 rasters the ratio is identical either way, so there's no compression-ratio win to be had here regardless.)

Repro:

```python
import os, base64, resource, zstandard as z
# run once per `threads` value in a fresh process so maxRSS doesn't bleed
raw = base64.b64encode(os.urandom(int(24e6 * 0.75))).encode()
rss = lambda: resource.getrusage(resource.RUSAGE_SELF).ru_maxrss / 1024**2  # macOS bytes
b = rss(); out = z.ZstdCompressor(threads=THREADS).compress(raw); print(rss() - b, "MB", len(out))
```

PR #466, opened 2026-06-23
add optimized path for image text that avoids re-parsing whole document

bd-mkt:
# perf: cheap picture text at ingest, gated by `processing.fast_picture_text`

## Summary

`extract_item_text` rendered every description-less `PictureItem` through
`PictureItem.export_to_markdown`, which constructs a fresh whole-document
`MarkdownDocSerializer` per picture. In the deployed docling-core, building that
serializer re-runs `DoclingDocument`'s `@model_validator(mode="after")` hooks —
including the bbox clamping that walks every table cell on every page. Called
once per picture (in `build_picture_chunks`) and again per item (in
`extract_items`), this is O(pictures × whole-document) and dominated ingest time
on image-heavy documents.

This PR makes the description-less picture path use the cheap
`FloatingItem.caption_text` accessor (O(captions), no serializer, no
re-validation), gated behind a new `processing.fast_picture_text` flag that
defaults to the fast path. Setting it to `false` restores the exact
`export_to_markdown` text.

## Evidence

Production thread dump showed the hot path:

```
build_picture_chunks → extract_item_text → export_to_markdown
  → pydantic __init__ → validate_document
  → _clamp_provenance_bboxes_to_pages → clamp_table_cell_bboxes → _clamp_bbox_to_page
```

i.e. a full-document validation per picture, producing text that — with an empty
image placeholder and no description — is just the caption anyway.

## Changes

- **`config/models.py`** — added `ProcessingConfig.fast_picture_text: bool = True`
  with docs noting it governs ingestion only; rebuild and store migrations
  always use the fast path.
- **`store/models/document_item.py`** — `extract_item_text(..., *,
  fast_picture_text=True)`. The VLM-description check still runs first and is
  unchanged; the flag only chooses the description-less fallback
  (`caption_text` vs `export_to_markdown`). `extract_items(..., *,
  fast_picture_text=True)` forwards it.
- **`client/processing.py`** — `build_picture_chunks` and `_merge_picture_chunks`
  forward the flag; `chunk()` reads `config.processing.fast_picture_text`.
- **`client/documents.py`** — the three ingestion `extract_items` call sites pass
  `client._config.processing.fast_picture_text`.
- **`tests/test_document.py`** — `test_extract_item_text_picture_modes`: the slow
  path equals the placeholder export; the fast path returns the caption and, via
  a monkeypatched `export_to_markdown` that raises, proves no serializer is built.

## Behavior

- **Default (`true`):** description-less picture text comes from `caption_text`.
  For caption-less, description-less pictures (the common image-heavy case) this
  is byte-identical (`""`) to the old export, just far cheaper. For captioned
  pictures it returns the caption text; it does not include non-caption
  annotation prose or markdown formatting that `export_to_markdown` would.
- **`false`:** exact prior `export_to_markdown` output, with the per-picture
  full-document re-validation cost.
- **VLM descriptions:** unaffected. `_picture_description_text` is checked before
  the flag, so a picture with a description returns it verbatim regardless of the
  flag; the flag only governs the fallback for pictures the VLM did not describe.
- **Scope:** rebuild (`rebuild.py`) and migrations (`v0_40_0`, `v0_48_0`) are not
  threaded with config and use the fast default.

## Testing

- `uv run pytest tests/test_document.py tests/test_picture_in_context.py` passes
  (incl. the new test).
- `ruff check` clean on all changed files.

## Follow-up (not in this PR)

The `TableItem` branch of `extract_item_text` has the same per-call serializer
cost but can't use a cheap accessor (the markdown *is* the content); a
build-the-serializer-once approach in `extract_items` would address it if tables
prove to be a hotspot. The new config field is not yet reflected in any config
reference doc / example YAML.

### DIFF
--- a/haiku_rag_slim/haiku/rag/client/documents.py
+++ b/haiku_rag_slim/haiku/rag/client/documents.py
@@ -118,7 +118,12 @@ async def _store_document_with_chunks(
     Handles versioning/rollback on failure.
     """
     chunks = await ensure_chunks_embedded(client._config, chunks, client.embedder)
-    items = await asyncio.to_thread(extract_items, "", docling_document)
+    items = await asyncio.to_thread(
+        extract_items,
+        "",
+        docling_document,
+        fast_picture_text=client._config.processing.fast_picture_text,
+    )
 
     async with client.store._write_lock:
         versions = await client.store.current_table_versions()
@@ -175,7 +180,11 @@ async def _update_document_with_chunks(
     items: list[DocumentItem] | None = None
     if docling_document is not None:
         items = await asyncio.to_thread(
-            extract_items, document.id, docling_document, existing_picture_data
+            extract_items,
+            document.id,
+            docling_document,
+            existing_picture_data,
+            fast_picture_text=client._config.processing.fast_picture_text,
         )
 
     async with client.store._write_lock:
@@ -282,8 +291,13 @@ async def _store_documents_with_chunks(
         for _, chunks, _ in prepared
     ]
 
+    fast_picture_text = client._config.processing.fast_picture_text
+
     def _extract_all_items():
-        return [extract_items("", d) for _, _, d in prepared]
+        return [
+            extract_items("", d, fast_picture_text=fast_picture_text)
+            for _, _, d in prepared
+        ]
 
     all_item_lists = await asyncio.to_thread(_extract_all_items)
 
--- a/haiku_rag_slim/haiku/rag/client/processing.py
+++ b/haiku_rag_slim/haiku/rag/client/processing.py
@@ -173,11 +173,13 @@ def _merge_picture_chunks(
     text_chunks: list[Chunk],
     document_id: str | None,
     existing_picture_data: dict[str, bytes] | None,
+    fast_picture_text: bool,
 ) -> list[Chunk]:
     picture_chunks = build_picture_chunks(
         docling_document,
         document_id=document_id,
         existing_picture_data=existing_picture_data,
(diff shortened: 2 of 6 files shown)


ggozad (commented):
Thanks for profiling this properly! I'd land it differently:

1. **Drop the `fast_picture_text` flag.** For pictures our converters produce the two paths are equivalent: a described picture returns `meta.description` before the flag us read, and a description-less picture has no legacy annotations (we never set `do_picture_classification` / `do_chart_extraction`), so the export is just the caption. They only diverge for an externally-enriched doc passed to `import_document()` which is too narrow for a global knob.

2. **Space-join captions instead of `caption_text`.** `caption_text` concatenates with no separator, so a 2-caption picture becomes `"FirstSecond"` vs the serializer's `"First Second"`, which merges tokens for FTS/embeddings, no annotations involved.

3. **Tables have the same cost.** `TableItem.export_to_markdown(doc)` is just `MarkdownDocSerializer(doc=doc).serialize(item=self).text`, so the table branch also builds a serializer per table. Reusing one lazily-created serializer per `extract_items()` call is byte-identical output, 4× `validate_tree` instead of 4×N.

The picture half is three lines in `extract_item_text` with no flag or config:

```python
if isinstance(item, PictureItem):
    if description := _picture_description_text(item):
        return description
    return " ".join(c.resolve(docling_doc).text for c in item.captions)
```

I'll put up a separate PR covering pictures & tables, no flag, table output unchanged

ggozad:
Superseded by #467

PR #471, opened 2026-06-25
re-constitute split pdfs on the fly rather than holding extra copies

bd-mkt:
# perf: merge PDF slices incrementally to halve conversion peak memory

## Summary

`convert_pdf_with_splitting` converted every PDF slice, **collected all of them
in a list**, then called `DoclingDocument.concatenate` at the end.
`concatenate` deep-copies every item of every input into an internal
`_DocIndex` merge buffer, so the collect-then-concatenate shape holds **all
slices plus the merged copy at once** — roughly twice the document in memory at
peak, for large image-bearing PDFs.

This PR drives the same `_DocIndex` **incrementally**: each converted slice is
folded into a running merge buffer and released before the next slice is
converted. Peak working set drops to ~one merged document plus one slice
(roughly halved), at identical total CPU. The merged result is byte-for-byte
identical to `concatenate`, because the incremental path issues the exact same
sequence of `_DocIndex.index()` calls — just interleaved with conversion.

## Why not naive folding

Repeatedly calling `concatenate([accumulator, next_slice])` was rejected: each
call re-deep-copies the entire growing accumulator (O(n²) CPU) and the final
fold still rebuilds the whole document, so peak memory is unchanged. The win
requires a *single* running buffer indexed once per slice — exactly what
`concatenate` does internally, just not exposed for incremental use.

## Changes

- **`converters/pdf_split.py`**
  - `_incremental_merge_available()` — probes for docling-core's internal
    `_DocIndex` / `_update_from_index`.
  - `_SliceMerger` — `add()` folds one slice into the running `_DocIndex`
    (deep-copies its items, then the slice is dropped); `result()` materialises
    the merged document. Falls back to collecting slices and calling the public
    `concatenate()` when the internals are unavailable (version drift).
  - `convert_pdf_with_splitting` now folds each converted slice via
    `await asyncio.to_thread(merger.add, slice_doc)` and `del`s it before the
    next slice converts, instead of accumulating all slices. Slice-failure abort
    semantics and the pdfium-locked generator cleanup are unchanged.
- **`tests/test_pdf_split.py`**
  - `test_incremental_merge_matches_concatenate` — asserts the incremental
    merger is byte-identical (`model_dump_json()`) to `concatenate`, and asserts
    the incremental path is active (canary for docling internals drift).
  - `test_slice_merger_falls_back_to_concatenate` — forces the capability probe
    off and verifies the fallback still matches `concatenate`.
  - Replaced `test_concatenate_runs_off_event_loop_thread` with
    `test_slice_merge_runs_off_event_loop_thread` — the old test monkeypatched
    `concatenate`, which the incremental path no longer calls; the new one spies
    `_SliceMerger.add` and asserts every fold runs off the event-loop thread.

## Testing

- `tests/test_pdf_split.py` — 9 passed (incl. equivalence + fallback).
- Real-PDF integration `test_split_and_merge_matches_single_pass` (docling-local
  over the 9-page DocLayNet paper) — passed: merged output is byte-identical
  markdown to single-pass conversion, confirming the incremental merge matches
  prior behavior on real conversions.

## Effect and limits

- Conversion-phase peak per in-flight document drops from ~2x the merged
  document to ~1x + one slice. Same total CPU (each item deep-copied once; no
  O(n²) folding penalty).
- **Floor unchanged:** the fully merged `DoclingDocument` must still exist for
  chunk -> embed -> compress, so per-slot memory still bottoms out at ~1x the
  document. Going lower would require per-slice chunk/embed/store — a separate,
  larger re-architecture.
- This stacks with the separate `claim_timeout_s` change that removed the
  reaper-induced duplicate-copy convoy.

## Risk

`_SliceMerger` reaches into docling-core internals (`_DocIndex`,
`_update_from_index`). Mitigated by: the runtime fallback to public
`concatenate`, and the `_incremental` canary assertion in the equivalence test,
which fails loudly in CI if a docling upgrade changes those internals rather
than silently corrupting merges.

### DIFF
--- a/haiku_rag_slim/haiku/rag/converters/pdf_split.py
+++ b/haiku_rag_slim/haiku/rag/converters/pdf_split.py
@@ -71,6 +71,63 @@ def iter_pdf_slices(
             src.close()
 
 
+def _incremental_merge_available() -> bool:
+    """Whether docling-core exposes the internal merge buffer this module
+    drives incrementally. False on versions that renamed/removed it, in which
+    case ``_SliceMerger`` falls back to the public ``concatenate``."""
+    from docling_core.types.doc.document import DoclingDocument
+
+    return hasattr(DoclingDocument, "_DocIndex") and hasattr(
+        DoclingDocument, "_update_from_index"
+    )
+
+
+class _SliceMerger:
+    """Merges converted PDF-slice documents into one ``DoclingDocument``.
+
+    ``DoclingDocument.concatenate`` indexes each input into an internal
+    ``_DocIndex`` merge buffer (deep-copying its items), then materialises the
+    result. Collecting every slice up front and concatenating at the end thus
+    holds all slices *and* the merged copy at once — roughly twice the
+    document. This merger drives the same ``_DocIndex`` incrementally: each
+    slice is folded in and released before the next is converted, so peak
+    working set is ~one merged document plus one slice. The sequence of
+    ``index()`` calls is identical to ``concatenate``'s, so the merged result
+    is byte-for-byte the same.
+
+    ``_DocIndex``/``_update_from_index`` are docling-core internals; when they
+    are unavailable (version drift, per ``_incremental_merge_available``) the
+    merger collects slices and calls the public ``concatenate`` instead.
+    """
+
+    def __init__(self) -> None:
+        from docling_core.types.doc.document import DoclingDocument
+
+        self._incremental = _incremental_merge_available()
+        self._index = DoclingDocument._DocIndex() if self._incremental else None
+        self._collected: list[DoclingDocument] = []
+
+    def add(self, slice_doc: "DoclingDocument") -> None:
+        """Fold one slice into the merge. CPU-bound (deep-copies the slice's
+        items), so call via ``asyncio.to_thread`` to keep it off the loop."""
+        if self._incremental:
+            assert self._index is not None
+            self._index.index(slice_doc)
+        else:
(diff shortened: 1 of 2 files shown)


ggozad:
I am not sure the 2x holds for image-bearing PDFs. After conversion, especially via docling-serve, which uses `model_validate`, page/picture rasters live as base64 `data:` strings in `ImageRef.uri` and `_pil` is `None`. `concatenate`'s internal `copy.deepcopy` *shares* that immutable string buffer rather than copying it. So collecting all slices and then concatenating never holds two copies of the rasters. Only the structural items (text/bbox/cells) get duplicated, and those are tiny next to the rasters.
I could be wrong here, but that's my first take...

ggozad:
Closing this. I re-measured on docling-core 2.82.0 with serve-style slices (40 pages, ~106 MB of base64 rasters): collect-then-concatenate peaked at 320.9 MB RSS vs 322.5 MB for the incremental _DocIndex merge — 1.00 ratio, not the ~0.5 assumed. Rasters live as base64 strings in ImageRef.uri (Rust-backed AnyUrl), and concatenate's internal copy.deepcopy shares that buffer instead of copying (deepcopy of a ~13 MB doc adds 0.0 MB RSS), so collecting all slices never holds a second copy and the merged doc references every page's raster regardless. The incremental path only avoids duplicating structural items (text/bbox/cells), negligible next to rasters and only relevant for text-only docs that don't OOM.

PR #472, opened 2026-06-25
 resilience: retry docling-serve requests on another instance when one crashes

bd-mkt:
# resilience: retry docling-serve requests on another instance when one crashes

## Summary

docling-serve instances crash under memory leaks (some documents trigger them
more often). Today a request that lands on a crashing/unresponsive instance
fails the whole document: the error propagates to the ingester pipeline,
classified `TransientError`, and the queue re-converts the *entire* document —
which may hit another crash, eventually DLQ'ing after `max_attempts` and burning
the fleet.

This PR makes `DoclingServeClient` retry a failed request on a **different**
instance before giving up. A single instance dying mid-request now fails over to
a healthy one transparently, instead of failing the document. It covers both
sliced (`split_pages`) and single-pass conversions, since it lives at the client
layer.

## Changes

`providers/docling_serve.py`:

- **`_is_retryable(exc)`** — retries only "instance unhealthy" signals:
  `httpx.TransportError` (connection reset / timeout — a crashed or hung
  instance) and HTTP 5xx / 408 / 429. Other 4xx and the task-level `ValueError`
  raised by `_submit_and_wait` (task failure / missing task_id) are **not**
  retried — they won't succeed elsewhere, and retrying a deterministically bad
  document would just burn the fleet (that case is left to upstream
  classification / a future quarantine).
- **`_pick_url(exclude=…)`** — round-robins as before, but a retry prefers an
  instance that hasn't already failed *this* request. Falls back to cycling when
  every instance is excluded, so a single-instance fleet still retries after
  backoff (the instance may have been restarted by the orchestrator).
- **`_run_with_retry(attempt, name)`** — runs the full submit → poll → fetch
  trio against one instance (task IDs are instance-local, so a trio can't be
  split), retrying up to `max_attempts` with capped exponential backoff and
  failover. Non-retryable errors propagate immediately; each failover logs a
  warning.
- `submit_and_poll` / `submit_and_poll_zip` now run their attempt through the
  wrapper. The `files` payload is bytes (the converter reads the file fully into
  memory), so re-sending it across retries is safe.
- New constructor params `max_attempts` (default 3), `retry_base_delay`,
  `retry_max_delay`. Not yet wired to `providers.docling_serve` config — a
  trivial follow-up if ops tuning is wanted.

## Behavior

- Transport error or 5xx on instance A → retry on instance B (preferring
  not-yet-failed instances) → success, error never surfaces.
- Every instance down → retries up to `max_attempts`, then raises the transport
  error (→ pipeline `TransientError` → whole-document retry, as before).
- 4xx (other than 408/429) → single attempt, propagates immediately (no
  pointless failover).
- Normal success path is unchanged: one attempt, one `_pick_url()`, so the
  process-global round-robin distribution across the fleet is preserved.

## Scope / non-goals

- This is failover *reactively* after a failure. It does **not** proactively
  route around a known-dead instance — a crashed instance is still tried first
  ~1/fleet-size of the time and absorbed by failover. Proactive avoidance is a
  follow-up (per-instance circuit breaking).
- Does not change slice-merge or document-level retry semantics.

## Testing

- `tests/test_docling_serve_client.py` — 11 passed (7 existing + 4 new):
  failover on connection error, failover on 5xx, exhaustion-then-raise across
  all-down instances, and no-retry-on-4xx (asserts a single attempt). Existing
  round-robin / lifecycle-pinning tests still pass, confirming the success-path
  distribution is unchanged.

Note: three `TestDoclingServeConverter` tests (`test_convert_file_text`,
`test_convert_file_pdf`, `test_parse_zip_leaves_unknown_artifact_uri_unchanged`)
fail on Windows with a `PermissionError` reopening a `NamedTemporaryFile` — a
pre-existing environment issue, confirmed failing identically on the baseline
without this change; they pass on Linux/CI.

### DIFF
--- a/haiku_rag_slim/haiku/rag/chunkers/docling_serve.py
+++ b/haiku_rag_slim/haiku/rag/chunkers/docling_serve.py
@@ -56,9 +56,15 @@ class DoclingServeChunker(DocumentChunker):
 
     def __init__(self, config: AppConfig = Config):
         self.config = config
+        ds = config.providers.docling_serve
         self.client = DoclingServeClient(
-            base_urls=config.providers.docling_serve.base_urls,
-            api_key=config.providers.docling_serve.api_key,
+            base_urls=ds.base_urls,
+            api_key=ds.api_key,
+            max_attempts=ds.max_attempts,
+            retry_base_delay=ds.retry_base_delay,
+            retry_max_delay=ds.retry_max_delay,
+            breaker_failure_threshold=ds.breaker_failure_threshold,
+            breaker_cooldown_s=ds.breaker_cooldown_s,
         )
         self.chunker_type = config.processing.chunker_type
 
--- a/haiku_rag_slim/haiku/rag/config/models.py
+++ b/haiku_rag_slim/haiku/rag/config/models.py
@@ -227,6 +227,32 @@ class DoclingServeConfig(BaseModel):
 
     base_url: str | list[str] = "http://localhost:5001"
     api_key: str = ""
+    max_attempts: int = Field(
+        default=3,
+        ge=1,
+        description="Total attempts per request before giving up; each retry "
+        "fails over to another instance (instances crash on memory leaks).",
+    )
+    retry_base_delay: float = Field(
+        default=0.5,
+        ge=0,
+        description="Base seconds for the exponential backoff between retries.",
+    )
+    retry_max_delay: float = Field(
+        default=8.0, ge=0, description="Cap on the retry backoff, in seconds."
+    )
+    breaker_failure_threshold: int = Field(
+        default=3,
+        ge=1,
+        description="Consecutive failures before an instance's circuit breaker "
+        "opens and routing skips it.",
+    )
+    breaker_cooldown_s: float = Field(
+        default=30.0,
+        ge=0,
+        description="How long an open breaker skips an instance before allowing "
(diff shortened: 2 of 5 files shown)


ggozad:
This and #473 (per-instance circuit breaking) are similar and touch the same files.

- Both rewrite `_pick_url` and the submit/poll bodies. The combined `_pick_url`
  needs one signature that *both* takes `exclude=…` (this PR) and skips
  breaker-open instances (#473).
- `_is_retryable` (this PR) and `_is_instance_failure` (#473) are the identical
  predicate — TransportError + 5xx/408/429. There should be one helper, not two
  names for the same logic.
- The breaker bookkeeping #473 puts in `submit_and_poll` belongs inside
  `_run_with_retry` (record success/failure per attempt), so failover and the
  breaker live in one place.

They should land them as one PR (or stack #472 → #473)

ggozad:
`max_attempts` / `retry_base_delay` / `retry_max_delay` are hardcoded constructor
defaults, unreachable from config. Add them to `DoclingServeConfig` and pass
through `get_converter`/`get_chunker`.

bd-mkt:
implemented updates

ggozad:
Consolidated in https://github.com/ggozad/haiku.rag/pull/486

PR #473, opened 2026-06-25
change breakers to be per-instance

bd-mkt:
# resilience: per-instance circuit breaking for docling-serve

## Summary

docling-serve instances crash under memory leaks. The client round-robins
requests across the fleet, but the rotation was **health-blind** — it kept
handing requests to an instance that had already crashed, so ~1/fleet-size of
all requests failed against a known-dead node until something external removed
it.

This PR adds a per-instance circuit breaker. After an instance fails enough
times, its breaker opens and `_pick_url` skips it; after a cooldown the breaker
auto-probes, and a successful request closes it. The fleet now routes around a
crashed/leaking instance instead of repeatedly hitting it.

## Changes

`providers/docling_serve.py`:

- **`_InstanceBreaker`** — a small, self-contained closed/open-with-cooldown
  breaker. Kept local rather than reusing the ingester's `CircuitBreaker` so the
  provider layer doesn't take a backwards dependency on the ingester package
  (whose `pollers/__init__` would also pull in a circular import).
- **`_instance_breakers`** — process-global registry keyed by base URL (mirrors
  the existing `_instance_rotators`), so an instance that crashed stays skipped
  across the per-job clients until its cooldown elapses.
- **`_is_instance_failure(exc)`** — only transport errors (connection reset /
  timeout) and HTTP 5xx / 408 / 429 count against instance health. Other 4xx
  (caller's fault) and the task-level `ValueError` from `_submit_and_wait`
  (document problem) do not trip the breaker.
- **`_pick_url()`** — skips instances whose breaker is open, falling back to the
  first-seen instance when every breaker is open (a single-instance or
  fully-down fleet still gets a probe rather than no attempt).
- **`submit_and_poll` / `submit_and_poll_zip`** — record success/failure per
  request via `_record_outcome`, which trips the breaker only on instance-health
  failures and logs a warning when a breaker opens.
- New constructor params: `breaker_failure_threshold` (default 3),
  `breaker_cooldown_s` (default 30s), and an injectable `now_fn` (for tests).
  Not yet wired to `providers.docling_serve` config — a trivial follow-up.

## Behavior

- Instance fails `threshold` times → breaker opens → subsequent `_pick_url`
  calls skip it for `cooldown_s`.
- After the cooldown the breaker allows one probe; success closes it, another
  failure re-opens it.
- 4xx and task-level errors leave the breaker untouched.
- All-open fleet / single-instance fleet still gets a probe (no hard failure
  from "nothing to pick").
- Success path is otherwise unchanged: all-closed breakers behave exactly like
  the prior round-robin (existing distribution / lifecycle-pinning tests pass
  unchanged).

## Scope / relationship to retry+failover

This is built on the current baseline; the reactive retry-on-another-instance
change (in-request failover) is a separate PR. The two compose: this breaker
records outcomes and skips dead instances at pick time, while the retry PR adds
per-request failover and a `_pick_url(exclude=...)` argument. The only merge
point is `_pick_url`'s signature (combine breaker-skipping with the per-request
`exclude` set) — straightforward.

## Testing

- `tests/test_docling_serve_client.py` — 10 passed (7 existing + 3 new):
  - open breaker skips the crashed instance (later requests never attempt it);
  - breaker recovers after cooldown (injected clock) and a success closes it;
  - a 4xx does not trip the breaker even at a 1-failure threshold.
  - Existing round-robin / lifecycle-pinning tests still pass, confirming the
    all-healthy success path is unchanged.

Note: three `TestDoclingServeConverter` tests (`test_convert_file_text`,
`test_convert_file_pdf`, `test_parse_zip_leaves_unknown_artifact_uri_unchanged`)
fail on Windows with a `PermissionError` reopening a `NamedTemporaryFile` — a
pre-existing environment issue, confirmed failing identically on the baseline
without this change; they pass on Linux/CI.

### DIFF
--- a/haiku_rag_slim/haiku/rag/chunkers/docling_serve.py
+++ b/haiku_rag_slim/haiku/rag/chunkers/docling_serve.py
@@ -56,9 +56,11 @@ class DoclingServeChunker(DocumentChunker):
 
     def __init__(self, config: AppConfig = Config):
         self.config = config
+        ds = config.providers.docling_serve
         self.client = DoclingServeClient(
-            base_urls=config.providers.docling_serve.base_urls,
-            api_key=config.providers.docling_serve.api_key,
+            base_urls=ds.base_urls,
+            api_key=ds.api_key,
+            breaker_config=ds.circuit_breaker,
         )
         self.chunker_type = config.processing.chunker_type
 
--- a/haiku_rag_slim/haiku/rag/circuit_breaker.py
+++ b/haiku_rag_slim/haiku/rag/circuit_breaker.py
@@ -5,7 +5,11 @@ from haiku.rag.config import CircuitBreakerConfig
 
 
 class CircuitBreaker:
-    """Three-state breaker over discover() failures.
+    """Three-state circuit breaker (closed / open with cooldown).
+
+    Shared by the ingester's per-source discover() breaker and the
+    docling-serve per-instance breaker — lives here, below both
+    ``providers`` and ``ingester``, so neither has to depend on the other.
 
     - closed: failures are counted; threshold flips to open.
     - open: probes are blocked until cooldown elapses, then a single probe
--- a/haiku_rag_slim/haiku/rag/config/models.py
+++ b/haiku_rag_slim/haiku/rag/config/models.py
@@ -218,6 +218,20 @@ class OllamaConfig(BaseModel):
     )
 
 
+class CircuitBreakerConfig(BaseModel):
+    """Three-state breaker (closed -> open -> probe) over consecutive failures.
+    Used by the ingester's per-source discover() breaker and the docling-serve
+    per-instance breaker."""
+
+    failure_threshold: int = Field(
+        default=5, description="Consecutive failures before the breaker opens."
+    )
+    cooldown_s: float = Field(
+        default=600.0,
+        description="How long the breaker stays open before allowing a probe.",
+    )
+
+
 class DoclingServeConfig(BaseModel):
     """docling-serve endpoints. Accepts a single URL or a list — when a list
     is given, the client round-robins jobs across the URLs. Each job's
@@ -227,6 +241,14 @@ class DoclingServeConfig(BaseModel):
 
     base_url: str | list[str] = "http://localhost:5001"
     api_key: str = ""
+    circuit_breaker: CircuitBreakerConfig = Field(
+        default_factory=lambda: CircuitBreakerConfig(
+            failure_threshold=3, cooldown_s=30.0
+        ),
+        description="Per-instance circuit breaker: skip an instance after this "
+        "many consecutive failures for this cooldown. Defaults are tuned faster "
+        "than the discover breaker since instances restart quickly.",
+    )
 
     @property
     def base_urls(self) -> list[str]:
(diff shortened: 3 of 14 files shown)


ggozad (changes requested):
`_InstanceBreaker` is a near-verbatim copy of `ingester/pollers/circuit_breaker.py:CircuitBreaker` (same state, `is_open`, `record_success`/`record_failure`, `now_fn`).

The circular-import justification doesn't hold: `ingester/pollers/__init__` doesn't transitively import `providers.docling_serve`, so there's no runtime cycle. 

Sketch:
- Move `CircuitBreaker` to a neutral `haiku/rag/circuit_breaker.py` (deps are only `time` + `CircuitBreakerConfig`, both below `providers`/`ingester`). Repoint the four ingester import sites + two tests. No behavior change.
- Drop `_InstanceBreaker`; have `_breaker_for` construct `CircuitBreaker` with a per-instance `CircuitBreakerConfig(failure_threshold=..., cooldown_s=...)` and `now_fn`. Everything else unchanged.

ggozad (changes requested):
When every breaker is open, `_pick_url` consumes exactly one full rotation, so the cursor lands back where it started and the fallback is the same URL every call. Repro (3 open instances): six picks all return `http://all-open-a:5001`. For the all-429 overload case this piles retries on one node.

Fix: on the all-open path, `return next(self._instance_rotator)` (one extra step) so the cursor advances by N+1 and the fallback rotates a → b → c. Add a test asserting six picks over three open breakers cycle the URLs instead of pinning one.

ggozad (changes requested):
`breaker_failure_threshold=3` / `breaker_cooldown_s=30.0` are hardcoded constructor defaults, unreachable from config. Add a `circuit_breaker: CircuitBreakerConfig` to `DoclingServeConfig` and pass it through `get_converter`/`get_chunker`, so the docling-serve breaker is tunable like the discover breaker already is.

ggozad:
Thanks for addressing all the issues.
The problem is now #472 went to a different direction than #473's review: it folds a
breaker back in using the **old inline `_InstanceBreaker` + flat `breaker_*`
config fields**, which #473's review replaced, and its `_pick_url`
reintroduces the all-open pinning bug #473 just fixed. So the two PRs are now two
overlapping breaker implementations and can't both merge.

Let's please land them as **one PR**: keep #473's branch as the base, rebase #472's
failover onto it, and drop #472's breaker duplication. Target below.

### Target end state (`providers/docling_serve.py`)

Keep from #473:
- Shared `CircuitBreaker` from `haiku/rag/circuit_breaker.py`.
- `circuit_breaker: CircuitBreakerConfig` on `DoclingServeConfig`, threaded
  through `get_converter` / `get_chunker`.

Add from #472:
- `_run_with_retry` + retry config `max_attempts` / `retry_base_delay` /
  `retry_max_delay` on `DoclingServeConfig`.

Drop from #472:
- the inline `_InstanceBreaker` class;
- the flat `breaker_failure_threshold` / `breaker_cooldown_s` config fields
  (use the nested `circuit_breaker` instead).

Reconcile the overlap:
- **One `_pick_url(exclude=...)`** that skips breaker-open instances, honors
  `exclude`, and on the all-open / all-excluded path takes the extra
  `next(self._instance_rotator)` step so consecutive calls rotate. #472's variant
  returns the first not-excluded URL and so re-pins the same node every call,
  that's a regression; #473's all-open fix has to survive the merge.
- **One health predicate.** `_is_retryable` (#472) and `_is_instance_failure`
  (#473) are the identical TransportError + 5xx/408/429 test. Keep one name.
- **Breaker bookkeeping inside `_run_with_retry`** — record failure on a
  retryable error, record success on completion — so failover and breaker live in
  one place. Collapse `_record_outcome` / `_record_failure` to one helper.

### Tests
Union of both PRs' suites, deduping the now-identical breaker tests
(`test_4xx_does_not_trip_breaker`, cooldown recovery, all-open rotation,
config-reaches-client). Add one test for the failover + breaker interaction
(a retryable failure both fails over *and* counts toward the breaker).

### Docstring fix (fold into this PR)
`DoclingServeConfig`'s docstring currently says "for true load balancing or
failover, put an LB in front." That's misleading for the default
`LocalOrchestrator` mode — an LB only works in docling-serve RQ mode (shared
Redis task state). For standalone instances the trio is instance-pinned and this
client does the failover/breaking. Please qualify it.

### CHANGELOG (one terse entry)
```
- docling-serve client retries a failed request on another instance and skips
  instances whose per-instance circuit breaker is open.
```

ggozad:
Consolidated in #486

PR #490, opened 2026-07-08
add de-duplication for image embedding and add http connection pool

bd-mkt:
# Dedupe picture embeddings and pool the embedding HTTP client

## Summary

Embedding a single large, image-heavy PDF took
over 3 hours in the `document.embed` step. The document expands to 28k
picture items backed by only 1,100 unique images, and the embedder was
re-embedding every occurrence over a fresh, unpooled HTTP connection per
call. These images are likely icons of various sorts. This PR makes two 
targeted changes to the embedding path:

1. **Deduplicate picture content** — identical image bytes are embedded once
   and the vector is reused for every chunk that shares them.
2. **Pool the HTTP client** — the embedder reuses one keep-alive
   `httpx.AsyncClient` (and its connection/DNS resolution) across all text,
   query, and image requests instead of opening a new client per call. The
   pool/timeout settings are configuration-driven and the pooling now extends
   to the openai and ollama flows as well, not just vLLM.

For the problem document this collapses ~28,847 image-embedding requests to
~1,109 and removes a per-request TCP connect + name resolution.

## Motivation

Investigation of the 3-hour `document.embed` span (10,819 s, clean close)
found:

- Picture chunks were embedded **serially, one HTTP request per image**, with
  no deduplication, even though the document repeats a small set of images
  (header/watermark/figures) across ~1,000 pages.
- Each request opened a brand-new `httpx.AsyncClient`, paying a fresh
  connect + DNS resolution of the service name (Docker embedded DNS) every
  call, with no keep-alive.

### `haiku/rag/config/models.py`
- New `EmbeddingHTTPConfig` (`timeout_s`, `max_connections`,
  `max_keepalive_connections`, `keepalive_expiry_s`), attached to
  `EmbeddingsConfig` as `http`. Defaults preserve historical behavior
  (60 s timeout, 16/16 connections, 300 s keepalive). Placed on
  `EmbeddingsConfig` (provider-agnostic, alongside `batch_size`) rather than
  `EmbeddingModelConfig` (model identity), so it applies to whichever provider
  is configured. Exported from `haiku.rag.config`.

### `haiku/rag/embeddings/__init__.py`
- `embed_chunks` now caches picture embeddings by a FIPS-safe content hash
  (`hashlib.sha256(data, usedforsecurity=False)`), embedding each distinct
  image once and reusing the vector for duplicates. Output order is preserved
  (one vector appended per chunk, in chunk order).
- New `build_http_client(EmbeddingHTTPConfig)` helper builds the pooled
  `httpx.AsyncClient` (`Timeout` + `Limits`) in one place.
- `get_embedder` builds a pooled client from `embeddings.http` and passes it as
  `http_client=` to the openai and ollama providers (extending pooling/keepalive
  to those flows), and into the vLLM embedder.
- `EmbedderWrapper` optionally owns that client and `aclose()` closes it; a
  no-op when the underlying SDK owns its own transport (voyageai / cohere /
  sentence-transformers).

### `haiku/rag/embeddings/vllm.py`
- `VLLMMultimodalEmbedder` holds one lazily-created, lock-guarded shared
  `httpx.AsyncClient` built via `build_http_client` from the injected
  `EmbeddingHTTPConfig`. Reused across every request (text, query, image)
  instead of per-call. The former module-level pool constants are removed.
- Added `aclose()` to close the pooled client; idempotent and safe to call
  when no request was ever made (the client is created lazily).
- `_post` uses the shared client; error mapping (connect / timeout / 401 /
  other HTTP) is unchanged.

### `haiku/rag/client/__init__.py`
- `HaikuRAG.__aexit__` awaits `embedder.aclose()` (best-effort, so it cannot
  mask an unwinding exception) before closing the store. The ingester enters
  the client via `async with HaikuRAG(...)`, so the pooled client is released
  on `run-batch` exit.

## Testing

- `test_embed_chunks_dedupes_identical_pictures` — duplicate images embed
  once and vectors map back to the correct chunks in order.
- `test_vllm_reuses_pooled_client` — one client is constructed and reused
  across multiple requests; `aclose()` closes it.
- `test_vllm_aclose_without_request_is_noop` — `aclose()` is safe before any
  request.
- `test_embeddings_http_config_defaults` — defaults preserve historical
  transport behavior.
- `test_vllm_client_built_from_http_config` — the pooled client's timeout and
  pool limits come from `EmbeddingHTTPConfig`.
- `test_openai_embedder_owns_and_closes_pooled_client` — the openai flow builds
  one pooled client, hands it to the provider, and `aclose()` closes it.
- `test_openai_embedder_forwards_base_url` — a configured `base_url` reaches the
  provider.
- `test_embed_chunks_non_bytes_picture_not_deduped` and
  `test_embedder_aclose_without_owned_client_is_noop` — cover the defensive
  branches (non-bytes picture data; base `aclose` with no owned client).
- Full `tests/test_embedder.py` suite passes (53 tests); `ruff check` and
  `ruff format --check` clean.
- Branch coverage: every line/branch this PR adds or changes is exercised.
  (The fork enforces no coverage gate. Remaining uncovered lines in the touched
  files are pre-existing code the PR does not modify — pydantic-ai passthroughs
  and the voyageai/cohere/sentence-transformers `get_embedder` branches.)

## Notes / follow-ups

- A separate PR will add **concurrent** chunk embedding (bounded semaphore +
  `TaskGroup`, order-preserving) on top of the pooled client, taking the
  problem document from minutes to seconds. It will add a `concurrency` knob to
  `EmbeddingsConfig`; `embeddings.http.max_connections` should be sized `>=` it
  so the pool never becomes the limiter.


ggozad (commented):
Thanks Bryan, real problems found here, but fixed differently in #491. Closing in favor of that.

- Image dedup: Deduping at embed time still writes 28k near-identical vectors into the index, bloating it and polluting search. #491 dedupes in `build_picture_chunks` instead, and adds a `processing.min_picture_size` filter (default 64px). Existing DBs heal with `rebuild --rechunk`.
- vLLM per-request `httpx.AsyncClient`: confirmed, fixed in #491 (one client per embedder + `aclose()`). Same pattern existed in the vLLM/Jina rerankers, fixed too.
- openai/ollama pooling: premise doesn't hold. pydantic-ai already creates one client per provider instance and reuses it. Injecting a 60s-timeout client would also regress pydantic-ai's 600s default, risking timeouts on 512-doc batches against slow servers.
- `EmbeddingHTTPConfig`: dropped. Embedding is serial, so pool-size knobs have no effect today. If concurrent embedding lands later, that PR brings its own config.

ggozad:
Superseded by #491

PR #520, opened 2026-07-28
Enforce the analysis code timeout per call

ggozad:
Two fixes to how the analysis sandbox enforces `analysis.code_timeout`.

- The VFS bridge blocks Monty's worker thread while a read is outstanding, so Monty never gets an interpreter step in which to check its duration budget. Code that reads document after document overran a 60s budget by minutes. The deadline is now checked before each read.
- Monty spends `max_duration_secs` across the session, not per call, and the session is reused so variables persist. It is now budgeted for `code_timeout * max_executions`; at the old per-call value the first slow execution starved every later one.


ggozad:
Superseded by #419. Both fixes moved there, adapted to the monty 0.0.19 session API: the read deadline in `_run_on_loop`, and the session budget of `code_timeout * max_executions`.

Testing on 0.0.19 also showed why the deadline belongs inside the read callback. Raising there answers the worker's suspension and keeps the session usable. Cancelling `feed_run` from outside wedges the session with a protocol `RuntimeError` instead.

PR #522, opened 2026-07-29
add deadlock mitigations for vacuum

bd-mkt:
# Bound `vacuum`/`optimize` with an optional timeout to prevent teardown hangs

## Problem

`Store.vacuum()` awaits `table.optimize(...)` with no time bound. If a
compaction stalls in lance-core (e.g. a slow/again-locked dataset, a network or
overlay filesystem that stops responding), the `await` never returns. Because
vacuum is reached from three places — an explicit `client.vacuum()`, the
throttled background `_schedule_vacuum()` task, and the close-time drain in
`HaikuRAG.__aexit__` → `_await_vacuum_tasks()` (which `await`s outstanding
background vacuums **and** runs a final `store.vacuum()`) — a single stuck
optimize can **wedge client teardown indefinitely**. `async with HaikuRAG(...)
as client: ...` then hangs on exit and the process never returns.

This surfaced when running an in-process vacuum after ingestion: the main event
loop parks in `select()` forever, waiting on the optimize future.

## Change

Add `StorageConfig.vacuum_timeout_seconds: float | None = None` and wrap the
optimize loop in `asyncio.timeout(...)`:

- **`None` (default) → unbounded**, i.e. no behavior change for existing users.
- A finite value converts a stuck pass into a **logged, skipped** pass
  (`"Vacuum timed out after …s; skipping this pass"`), handled alongside the
  existing `OSError` skip. Vacuum is best-effort maintenance, so a timeout is
  non-fatal — the caller's coroutine is freed and teardown proceeds.

Because all three entry points funnel through `Store.vacuum()`, this one bound
protects explicit vacuums, background vacuums, and `__aexit__` teardown.

## Files

- `haiku_rag_slim/haiku/rag/config/models.py` — new `vacuum_timeout_seconds`
  field on `StorageConfig`.
- `haiku_rag_slim/haiku/rag/store/engine.py` — `Store.vacuum()` wraps the
  per-table `optimize()` loop in
  `asyncio.timeout(self._config.storage.vacuum_timeout_seconds)`; new
  `except TimeoutError` logs + skips.
- `tests/test_vacuum_debounce.py` — `test_vacuum_times_out_and_skips` (a slow
  `optimize` is bounded, logged, returns without raising) and
  `test_vacuum_unbounded_when_timeout_none` (default runs to completion).

## Caveats

- **Best-effort cancellation.** `asyncio.timeout` frees the awaiting coroutine,
  but if the underlying lance/native future doesn't honor cancellation the
  operation may linger in the background — the guarantee is "the caller/teardown
  no longer hangs," not a forced abort of the native work.
- Opt-in by default (`None`), so it's a safe, non-breaking addition; set
  `storage.vacuum_timeout_seconds` to activate.

## Testing

`uv run pytest tests/test_vacuum_debounce.py tests/test_versioning.py tests/store/test_tags.py`
— new timeout tests pass, and the existing vacuum-on-close / tag-retention /
versioning paths are unaffected (42 passed). `ruff check` / `ruff format` clean.

### DIFF
--- a/haiku_rag_slim/haiku/rag/config/models.py
+++ b/haiku_rag_slim/haiku/rag/config/models.py
@@ -59,6 +59,11 @@ class StorageConfig(BaseModel):
     data_dir: Path = Field(default_factory=get_default_data_dir)
     auto_vacuum: bool = True
     vacuum_retention_seconds: int = 86400
+    # Upper bound (seconds) on a single vacuum/optimize pass. ``None`` leaves it
+    # unbounded (the historical behavior). A finite value converts a stuck
+    # compaction into a logged, skipped pass instead of an unbounded hang that
+    # would also wedge client teardown (which drains background vacuums).
+    vacuum_timeout_seconds: float | None = None
 
 
 class LanceDBConfig(BaseModel):
--- a/haiku_rag_slim/haiku/rag/store/engine.py
+++ b/haiku_rag_slim/haiku/rag/store/engine.py
@@ -571,14 +571,29 @@ class Store:
                 # Evaluate config at runtime to allow dynamic changes
                 if retention_seconds is None:
                     retention_seconds = self._config.storage.vacuum_retention_seconds
-                # Perform maintenance per table using optimize() with configurable retention
-                retention = timedelta(seconds=retention_seconds)
-                for table in self._tables().values():
-                    await table.optimize(
-                        cleanup_older_than=await self._tag_safe_retention(
-                            table, retention
+                # Bound the pass so a stuck compaction degrades to a logged,
+                # skipped pass rather than an unbounded await -- which would also
+                # wedge client teardown (__aexit__ drains background vacuums).
+                # timeout=None leaves it unbounded (historical behavior).
+                timeout = self._config.storage.vacuum_timeout_seconds
+                async with asyncio.timeout(timeout):
+                    # Perform maintenance per table using optimize() with configurable retention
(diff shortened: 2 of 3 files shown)


ggozad:
`Store.vacuum` calls `table.optimize()`, and lance's compaction bounds itself by row count
(~1M rows per fragment). A `documents` table of multi-MB docling blobs never reaches that,
so the fragment stays a compaction candidate forever and every vacuum after an append
re-merges all of it. Peak RSS runs about 5x the table's data files: 48.7 GB for an 8.8 GB
table, 67 GB for 19 GB. That is the `Killed`. Identical on lancedb 0.36.0, so it is not
version-specific.

Which makes a timeout the wrong instrument:

- The process is being OOM-killed, and a timeout cannot preempt the kernel.
- `asyncio.timeout` frees our coroutine but cannot cancel lance's native compaction, as your
  caveat notes. It then escapes the `_vacuum_lock`/`_write_lock` block while that compaction
  may still commit, which is what the write lock is there to prevent.
- Since the native work continues and `_schedule_vacuum` keeps firing, each timed-out pass
  leaves its allocation in place and the next stacks on top.

Filed upstream at lancedb/lancedb#2325, which needs byte-based compaction batching plus
`CompactionOptions` on `AsyncTable.optimize` (the async API exposes neither today). The RAM
requirement and the mitigations are now documented under "Vacuum Memory Requirements" in
`docs/configuration/storage.md`.

Issue #532, opened 2026-08-10
Make citation of evidence explicit and verifiable


ggozad:
## Problem

An answer built from retrieved evidence should say what it was built from. At
present a model can search, answer correctly, and register nothing — the answer is
then unattributable, and a UI has nothing to show. The instructions ask for a
citation on every answer, but nothing checks it, and several paths skip the check
entirely.

There is also no way for a model to say "nothing in the knowledge base supports
this". Staying silent and forgetting to cite look identical.

## Plan

**Citing is required for every answer, and an empty citation is a valid answer.**
A model that finds nothing relevant declares that explicitly rather than saying
nothing. An empty declaration is recorded as ungrounded, which is distinct from a
grounded answer and never counts as compliance.

**A declaration is recorded only after the cited ids resolve.** An attempt naming
ids that do not exist is not a citation.

**A declaration belongs to one question and to the evidence available when it was
made.** If the model gathers more evidence after citing, the declaration no longer
covers the answer and is no longer current. Repeated valid citations made in the same
request combine rather than replace each other, so citing twice in one request cannot
weaken what was already declared. A citation made after further evidence declares
afresh: carrying the earlier ids forward would claim they were redeclared after
evidence they never saw.

**Status is derived, not stored:** no declaration, grounded, or explicitly
ungrounded. This makes contradictory records impossible.

**Enforcement is opt-in and owned by one capability.** A host that wants
citations enforced registers a citation policy capability alongside the retrieval
capabilities. It makes one decision per question, reading what every retrieval
capability recorded, so two capabilities cannot each demand a citation for the same
answer. Without it, citations are still recorded and still validated; they are just
not required.

**When a model ends a question without citing, it is asked once to do so** — a
request to record what grounded the answer it already gave, not to change the
answer. If the citation tool is no longer available at that point, the run records
a grounding violation instead of asking for a tool that is gone.

## What this does and does not guarantee

It guarantees that validated evidence ids were declared for the question, after
the latest evidence *any* retrieval capability produced in it — strictly after,
since a citation made in the same request as the evidence has not seen it.

It does not guarantee that every piece of evidence an answer rests on was cited.
Nothing observable in a conversation can establish that, and the status names above
should be read with that limit in mind.

## Depends on

#531 — the evidence records, and the question and epoch identities this needs.

Goal and acceptance criteria: #530.

Issue #531, opened 2026-08-10
Build history compaction from recorded state


ggozad:
## Summary

History compaction should be generated from what a run recorded, not by parsing and
editing the transcript it already sent. It should also be something a host chooses,
which today it is not.

## Why the current approach is the problem

Today the trimming logic reads the rendered messages: it splits a tool result on a
separator, finds chunk ids with a regular expression, and filters image parts out of
message content. That makes several things impossible to get right:

- A picture's owner is inferred from which tool result it sits next to. When several
  tools' results arrive in the same request, that inference is wrong.
- The transcript identifies a picture only by its reference within a document
  (`#/pictures/0`), and those repeat across documents. Two documents' first pictures
  are indistinguishable.
- A label describing an image is a sibling string inside the same content list, so a
  filter that removes images keeps the text describing them.
- The only unit visible to a parser is one tool result, so a limit that applies to
  the whole conversation cannot be expressed.

It is also unconditional: there is currently no way to turn trimming off.

## Plan

**Retrieval and history rewriting are separated.** The RAG and analysis capabilities
retrieve evidence and validate citations, and nothing else. They never rewrite
conversation history.

**Compaction becomes an optional capability.** A host that wants it registers one:

```python
capabilities = [
    create_rag(db_path, config),
    create_analysis(db_path, config),
    create_evidence_compaction(config),   # optional
]
```

It **discovers** the evidence capabilities through the run's capability registry, so
no capability holds a reference to another and a host running one capability needs no
wiring change. Leaving it out is the way to turn compaction off, and the transcript is
then untouched. Registering two fails immediately, since each would rewrite the
history and claim the whole budget.

**Per-capability evidence records.** Each evidence capability records, in its own
state, which evidence it retrieved, which it cited, and when. Nothing is written by
two capabilities: the host's state is JSON storage, and a shared record would be
overwritten by whichever capability validated it last. The compactor merges what it
discovers into a transient view per request and stores nothing itself.

**Question identity and epoch are derived, not owned.** A question is identified by
the number of messages that existed when it arrived; an evidence outcome or a
citation records the message count at which it happened. Both are computed
identically by every participant from the same history, so nothing needs a shared
counter, and they survive a run that resumes an unfinished question. This relies on
the history being append-only, which it is.

**An evidence capsule.** For earlier questions, one generated summary of the evidence
that was cited, grouped by the question that cited it, each entry keeping its chunk id
so it can be cited again. Other earlier tool results become short receipts, so every
[...]

Issue #530, opened 2026-08-10
Multi-turn conversations need a proper evidence model


ggozad:
## Problem

A capability (RAG or analysis) attaches its search results to the message history
as tool output. In a single-question run that works well. Across a conversation it
does not: every question adds its results to the history, so requests grow turn
after turn, and answer quality falls as they do.

Trimming the history naively causes the opposite problem. Once earlier results are
removed, a follow-up that refers back to them has nothing to work with — and some
of that evidence cannot be fetched again. "In that diagram you showed me, what
colour is the supervisor box?" contains no terms a search can use, so if the figure
is gone the question becomes unanswerable.

We want conversations where a follow-up can rely on what was already found,
without the request growing without limit.

## What needs to hold

- **Evidence used in an answer stays available.** If the model cited something, a
  later question about it should still work, including for figures that cannot be
  retrieved by text search.
- **Earlier questions carry only the evidence that was used, when a host asks for
  that.** History compaction is opt-in: a host that enables it sees uncited evidence
  dropped and cited evidence rendered once, with its provenance. This is not a bound,
  and not a guaranteed reduction for every request — added headers and labels can
  make a short conversation whose evidence was entirely cited slightly larger.
  Retained evidence still grows with the conversation, and there is no configured
  limit. A host that does not enable it gets a transcript that is never rewritten.
  Retrieval alone must never rewrite conversation history.
- **Follow-ups can refer to earlier questions and answers,** and know which
  document each piece of evidence came from.
- **Other participants' content is never touched.** Pictures the user attached to
  their question, and output belonging to tools from other capabilities, must not be
  removed or altered. Tool results from several tools can arrive in the same
  request, so ownership has to be tracked rather than guessed from position.
- **A resumed run keeps its own evidence.** A run can continue an existing question
  rather than start a new one — deferred tool results, an interruption, a suspended
  response. Evidence gathered for the question in progress must not be treated as
  belonging to an earlier one.
- **Citations remain resolvable** for the whole conversation, so a UI can still
  render what an answer was based on.

## Scope

This issue is the goal and the acceptance criteria.

- Mechanism: #531 — optional history compaction built from recorded state.
- Citation behaviour: #532 — explicit and verifiable citation of evidence.

PR #535, opened 2026-08-11
Record cited evidence, and make history compaction an explicit capability

ggozad:
Capabilities record the evidence they retrieve and cite. Compacting the conversation
becomes a separate capability a host registers, instead of something every RAG run
does to every request.

## What changes

- `RAGState.evidence` / `AnalysisState.evidence` hold a `CapabilityEvidenceRecord`:
  which chunks the capability retrieved, which it cited, and in which question. A
  question is identified by the message count when it arrived; an outcome by the count
  when it happened. Both are derived from the conversation, so no counter is shared and
  a resumed question keeps its identity.
- `citation_status(records, question=...)` derives `missing` / `grounded` /
  `ungrounded` across capabilities. A declaration counts only if it followed the newest
  evidence of every capability in that question.
- `EvidenceCompactionCapability` replaces earlier questions' evidence on the request
  with one capsule of everything that was cited, grouped by the question that cited it,
  with cited page images re-attached. Every other earlier evidence return becomes a
  short receipt. Cite acknowledgements, other capabilities' output and user-attached
  pictures are untouched.
- Page images carry a line naming the search result they came from, and a machine tag
  so compaction recognises its own attachments rather than guessing from position.
- Requests only: `all_messages()` keeps everything the run gathered.

## Migration

`RAGCapability` and `AnalysisCapability` no longer rewrite requests. To keep earlier
questions compacted, register the new capability alongside them:

```python
from haiku.rag.capabilities.compaction import create_capability as compaction

agent = Agent(model, capabilities=[rag(db_path=...), compaction()])
```

The chat TUI and the example backend do this already. Single-shot paths
(`client.ask`, `client.analyze`, the MCP tools) do not, having nothing earlier to
compact.

Resuming a run — no prompt, deferred tool results, an unfinished history tail — now
requires the host to carry the capability state alongside the message history. Without
it the identity of the question in progress is unknowable, and the run raises instead
of treating it as a new question.

## Not included

No budget, cap or truncation: cited evidence is kept whole. This reduces what a request
carries; it does not bound it. Pruning beyond that is an integrator's policy, on the
wire only.

Citation enforcement is not here. Nothing requires a citation, and `cite` still rejects
an empty list, so `ungrounded` is not yet reachable from the model's side.

## Verification

1892 tests, coverage 100%, ruff and ty clean. Inspected 6-turn conversations on
Qwen3.6-35B against the OpenRAG Bench corpus: the visual back-reference is answered
from the retained figure and cited, and peak images per request fall from 12 — which
exceeded the server's limit and failed the run — to 7.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,8 @@
 - `haiku.rag.store` no longer re-exports `Store`; import it from `haiku.rag.store.engine`.
 - `docling-local` reuses one docling `DocumentConverter` per set of conversion options instead of building one per document, so local layout, table and OCR models are no longer loaded per document. Conversions through a shared converter are serialized.
 - Prior-question tool output is trimmed from the model request only; `all_messages()` retains what the run gathered.
-- A search result carrying page images no longer counts as a turn boundary; evidence retrieved earlier in the same turn survives.
+- Evidence retrieved for the current question is no longer trimmed mid-question, whether or not it carries page images.
+- `rag_cite` / `analysis_cite` returns are no longer replaced by the prior-question notice.
 
 ## [0.73.0] - 2026-08-06
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,11 +3,19 @@
 
 ### Added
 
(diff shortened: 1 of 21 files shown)


ggozad:
Implements #530 and #531. Citation enforcement (#532) follows separately.

PR #537, opened 2026-08-11
Stop eagerly importing Store from haiku.rag.store

### LINKED ISSUE #533: haiku-rag --help takes several seconds to start
When I run `haiku-rag --help`, it takes roughly 6-7 seconds to return. I investigated the cause using 
```python 
python -X importtime -c "import haiku.rag.cli"
```

Most of the time is spent importing lancedb, pyarrow, and pydantic_ai. Obviously, none of these is actually required for `--help`.  The same unnecessary imports happen for plain `haiku-rag` (no subcommand), `--version` and `init-config`

This seems to be caused by two things:

- `haiku/rag/store/__init__.py` imports `Store` from `.engine` and reexports it as part of `__all__`. So importing anything under haiku.rag.store.*, even something like the `SearchType` enum, also pulls in the full lancedb/pyarrow/pydantic_ai chain that `engine.py` depends on. Since `cli.py` imports `SearchType` solely to define the `--search-type` CLI option, it ends up paying the import cost for the rest of that chain as a result.

- `cli.py` imports `HaikuRAGApp` at module level rather than inside the functions that use it, so it gets loaded before argument parsing takes place, and before it's known whether the invoked command actually needs it.

I have tested a fix locally.  Removing the `Store` re-export from `store/__init__.py` (a quick search suggests nothing in the application or ingester code relies on that re-export, and only a few test files import it that way) and moving the `HaikuRAGApp` import into the few functions that actually need it. This brought --help down from ~6.5s to ~1.5s on my machine, with no adverse effects.

Are you interested in a PR for this sort of thing? If so, let me know and I'll submit one for review

lawrenceakka:
Closes #533 

Do not export `Store` from `haiku_rag_slim/haiku/rag/store/__init__.py` as part of `__all__`.  Instead, when `Store` is needed, import it from haiku.rag.store.engine directly. 

Defer `HaikuRAGApp`'s import in `cli.py` behind `TYPE_CHECKING`/lazy imports, avoiding unnecessary heavier imports at CLI startup.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 ### Fixed
 
 - `cross-encoder` reranking no longer ties the scores of strongly-relevant candidates, which left their order to the sort. Scores remain 0-1.
+- haiku-rag CLI startup no longer imports `lancedb`, `pyarrow` and `pydantic_ai`.
+- `haiku.rag.store` no longer re-exports `Store`; import it from `haiku.rag.store.engine`.
 
 ## [0.73.0] - 2026-08-06
 
--- a/haiku_rag_slim/haiku/rag/cli.py
+++ b/haiku_rag_slim/haiku/rag/cli.py
@@ -4,7 +4,7 @@ import sys
 import warnings
 from importlib.metadata import version
 from pathlib import Path
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
 import typer
 from dotenv import find_dotenv, load_dotenv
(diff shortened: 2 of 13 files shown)


ggozad (changes requested):
Thank you, this is great!

A few things before merge:

- `ruff check` fails on two of the test files, both `I001`. This should be caught is you run the pre-commit hook.
- Please add a CHANGELOG entry under `[Unreleased]` / `Fixed`, something like:
  "`haiku-rag` CLI startup no longer imports lancedb, pyarrow and pydantic_ai.
  `haiku.rag.store` no longer re-exports `Store`; import it from
  `haiku.rag.store.engine`."
- Could you add a guard test so this cannot regress again? A subprocess check is
  enough:
  `import haiku.rag.cli, sys; assert "lancedb" not in sys.modules`
- Would you mind doing the same for the ingester CLI in this PR?
  `haiku-ingester --help` is still ~3.4s: `ingester/cli.py` imports
  `haiku.rag.ingester.app` at module level, which pulls the client and lancedb
  before argument parsing. Same treatment as `cli.py`, and the guard test can
  cover both entry points.

lawrenceakka:
All done. If you want me to squash, let me know

ggozad:
Thank you!

PR #538, opened 2026-08-12
Citation policy: require a declaration, accept an empty one

ggozad:
Stacked on #535 — six commits on top of its tip. Merge that first, then rebase this
onto main.

## What changes

- **`rag_cite` / `analysis_cite` accept an empty `chunk_ids`**, recording a declaration
  with no refs that derives `ungrounded`. The call used to be rejected, so a model with
  nothing to cite could only stay silent, which is indistinguishable from forgetting.
- **Citing again cannot narrow a question.** Declarations merge until an evidence
  outcome follows the standing one, at any epoch; only newer evidence starts a
  declaration afresh.
- **The instructions lose their carve-outs.** Neither a refusal nor a corpus-level
  computation exempts the call; both cite an empty list.
- **`CitationPolicyCapability`** makes one enforcement decision per question, however
  many evidence capabilities are registered. When a question ends undeclared it asks
  once for the model to record what grounded the answer it already gave, without asking
  it to change that answer. Registering the capability is the only switch: without it,
  citations are still recorded and validated, just not required.
- **`DiscoveredEvidence` and discovery move to `haiku.rag.capabilities.evidence`**, so
  both optional capabilities share them instead of one importing from the other.
  `RAGCapabilityBase.cite_available` becomes public for the same reason.

## What is enforced

Every answer in a conversation that has something to declare: this question produced an
evidence outcome, **or** something was cited earlier. Cited evidence stays available to
later answers — in a capsule when a compactor is registered, in full when not — so a
follow-up reusing it is enforced despite searching nothing. That case is why the
capability exists.

Not enforced: an answer that cited an empty list, which is a declaration; and a
conversation with neither a current-question evidence outcome nor any earlier citation.

Consequences to weigh before enabling it:

- Once anything has been cited, later turns are enforced too, a greeting included. It
  satisfies the policy with an empty citation, costing one extra request. Distinguishing
  a greeting from an answer needs a heuristic, and there is no non-brittle one.
- A question whose cite tool was withdrawn by a spent request limit is recorded in
  `CitationPolicyState.violations` rather than pointed at a tool that is gone. So is one
  that finishes undeclared for any other reason.
- A question ends when its response carries no tool calls, or one of them names an
  output tool — a structured answer arrives that way. `end_strategy="early"` can end a
  run in a way no single response reveals; `after_run` records those rather than asking
  for a citation too late.
- Whether a question was already asked comes from the message history, keyed on
  `[haiku.rag/citation-redirect]`, so it survives a resumption and asks again if a
  redirect never reached the model.

## Not included

Grounding accuracy. `grounded` means validated chunk ids were declared after the
evidence, not that they support the sentence they are attached to: one observed answer
was correct while citing a different author's biography.

## Verification

1918 tests (`-m "not integration"`), coverage enforced at 100%, ruff and ty clean.

Checked end to end against the OpenRAG Bench corpus on Qwen3.6-35B: two arms of 37
questions differing only in whether the policy capability was registered, 29 of them
labelled unanswerable. Explicit ungrounded declarations rose from 23 to 26,
grounded answers to labelled-unanswerable questions fell from 4 to 2, answerable
questions remained 8 of 8 in both arms, and one redirect fired across the policy arm.
Every answer was read by hand; no invented grounding was observed. One run per arm, and
three labels turned out answerable from the corpus, so these are descriptive figures.


ggozad:
Implements #532. Stacked on #535, so its diff shows only the citation work — rebase onto main once #535 merges.

PR #543, opened 2026-08-13
feat: add search_filter to evaluations

Cwiesen:
This PR adds a search_filter option to evaluations, exposing haiku.rag search's filtering capability to evaluations. This would allow us to filter down a db if it happens to contain multiple corpora, but we only want to evaluate a subset of them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Changelog
 ## [Unreleased]
 
+### Added
+
+- `evaluations run --filter/-f CLAUSE`: SQL `WHERE` clause over document columns, applied to the retrieval benchmark's searches and to every capability search during QA. Recorded as `document_filter` in experiment metadata.
+
 ### Removed
 
 - `wix` evaluation dataset and its reference config `evaluations/configs/wix.yaml`.
--- a/docs/benchmarks.md
+++ b/docs/benchmarks.md
@@ -89,11 +89,11 @@ evaluations:
 
 ### Restricting the corpus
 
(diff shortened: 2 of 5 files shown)


ggozad (changes requested):
The search_filter threading looks right: resolved once in evaluate_dataset so both phases score the same subset, it reaches the analysis sandbox too, and it wires up the document_filter param that was already sitting unused in capability_runner. Two things need fixing before it lands, and neither is the feature itself:

- docs/benchmarks.md reverts f4d8e744 — the ORB tables go back to pre-release numbers (0.9798 → 0.9709, the two Gemma-4 QA rows) and the provenance paragraph recording analysis uncited cases at 26.3% → 4.9% is gone. A rebase won't fix it, the merge base is already main HEAD; easiest is git checkout main -- docs/benchmarks.md and re-add just the --filter bullet and the new section.
- db_source isn't a document column, so every example clause errors with Schema error: No field named db_source (valid: id, uri, title, metadata, created_at, updated_at). Same string in the --help text, which is what people will copy. mqf isn't a dataset key either — something like evaluations run orb_text --skip-db --filter "uri LIKE '%arxiv%'" works, and metadata is the field to use if corpus tagging is the intent.
- ty check fails on test_benchmark.py:698 — await_args is _Call | None. main is clean and the hook runs repo-wide. call_args[1]["document_filter"] type-checks and matches the rest of the file.
- ruff format wants two blank lines before resolve_search_filter. pre-commit run --files on the touched paths catches both this and the above; CI only runs ruff check.
- DatasetSpec.search_filter has no users, and it's what pulls in resolve_search_filter, the --filter "" rule and its tests. Fine to keep if a dataset is about to use it. Also, --filter → search_filter → document_filter → filter is four names for one value; document_filter is already the eval-side one and the metadata key outlives the rest.

Cwiesen:
OK, I've gone back through and made adjustments based on your feedback. I think I've fixed everything you touched on, but let me know and I'll make additional adjustments

ggozad (approved):
Make some additional changes you missed, merging.

PR #551, opened 2026-08-17
Expose all chunk metadata on search results and citations

### LINKED ISSUE #546: Custom chunk-level metadata is stored, but cannot be used by search
It is currently possible for a custom chunker (for example) to store additional metadata on a chunk.  The current code will store and retrieve it from the database without problem.  But the search and citation code will silently drop anything which is not already envisioned by the relevant pydantic model. User code cannot therefore attach additional chunk metadata, and then subsequently retrieve it after a chunk is returned from a search. 

This problem is already solved for document level metadata through a plain dict field called `document_meta`. 

The proposed solution is to add a custom field, say `extra_metadata` to each of `ChunkMetadata`, `SearchResult`, and
`Citation` in which this additional data can be stored, similar to FetchResult.extra_metadata: dict[str, str], eg:

```python
Class ChunkMetadata(BaseModel):
    """
    Structured metadata for a chunk, including DoclingDocument references.

    Attributes:
        doc_item_refs: JSON pointer references to DocItems in the parent DoclingDocument
                       (e.g., ["#/texts/5", "#/texts/6", "#/tables/0"])
        headings: Section heading hierarchy for this chunk
                  (e.g., ["Chapter 1", "Section 1.1"])
        labels: Semantic labels for each doc_item (e.g., ["paragraph", "table"])
        page_numbers: Page numbers where the chunk content appears
        extra_metadata: Arbitrary metadata added by a third-party converter or #<---------HERE
                        chunker
    """

    doc_item_refs: list[str] = []
    headings: list[str] | None = None
    labels: list[str] = []
    page_numbers: list[int] = []
    extra_metadata: dict = {}     #<------------HERE

```

I suppose an alternative would be for a custom chunker simply to set top-level keys directly on `chunk.metadata` instead of nesting everything under an `extra_metadata` key. But `pydantic` ignores unknown top-level keys unless the model is configured with extra="allow", so this would just be dropped. `ChunkMetadata` would need `model_config = ConfigDict(extra="allow")`, and then SearchResult and Citation would need the same.

Are you happy with this approach? If so, I'd be happy to submit a PR if you think it is worthwhile.

lawrenceakka:
A custom chunker can store arbitrary metadata on a `Chunk` object as key-value pairs within the `Chunk.metadata` dictionary. It is already stored with the chunk. This PR exposes that metadata to search and citation results, through `SearchResult.chunk_meta` and `Citation.chunk_meta`. For context-expanded results, the metadata is that of the anchor chunk.

Notes:

- chunk_meta *includes* the standard metadata values which are already exposed in typed form elsewhere via `chunk.get_chunk_metadata` (presently `headings`, `page_numbers`, and `labels`).  
- The MCP tool `search_documents` returns serialized `SearchResult` objects directly, and each of those includes the `chunk_meta` dict. There is therefore some duplication of the standard metadata fields.  If the custom key/value pairs in `chunk_meta` are large, each MCP result will also be large.

Closes #546

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Changed
 
 - `import_documents` embeds chunks across the whole batch in one pass instead of per document.
+- Raw chunk metadata is now exposed to search and citation results, through `SearchResult.chunk_meta` and `Citation.chunk_meta`. For context-expanded results, the metadata is that of the anchor chunk.
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,11 +5,12 @@
 
 - `evaluations run --filter/-f CLAUSE`: SQL `WHERE` clause over document columns, applied to the retrieval benchmark's searches and to every capability search during QA. Recorded as `document_filter` in experiment metadata.
 - `mtrag_clapnq` / `mtrag_clapnq_rewrite` / `mtrag_clapnq_live` / `mtrag_clapnq_live_uncompacted` evaluation datasets: multi-turn QA with gold-prefix and live-session conversation replay, live arms with and without `EvidenceCompactionCapability`, Recall@k/nDCG@k retrieval metrics, eligibility-aware citation scoring, refusal precision/recall, per-turn tool-traffic attributes, and `citation_status` / `turn_citation_status` eval attributes.
+- Raw chunk metadata is now exposed to search and citation results, through `SearchResult.chunk_meta` and `Citation.chunk_meta`. For context-expanded results, the metadata is that of the anchor chunk.
+
 
 ### Changed
 
 - `import_documents` embeds chunks across the whole batch in one pass instead of per document.
-- Raw chunk metadata is now exposed to search and citation results, through `SearchResult.chunk_meta` and `Citation.chunk_meta`. For context-expanded results, the metadata is that of the anchor chunk.
 
 ### Removed
 
--- a/docs/custom-pipelines.md
+++ b/docs/custom-pipelines.md
@@ -105,6 +105,8 @@ for chunk in chunks:
     print(f"Headings: {meta.headings}")
     print(f"Page numbers: {meta.page_numbers}")
     print(f"Labels: {meta.labels}")
+    # Access raw metadata (including headings, page_numbers and labels)
+    print(f"Raw metadata: {chunk.metadata}")
 ```
 
 Chunks are returned with:
@@ -114,6 +116,10 @@ Chunks are returned with:
 - `metadata` - Dict with `doc_item_refs`, `headings`, `labels`, `page_numbers`
 - `embedding` - `None` (not yet embedded)
 - `document_id` - `None` (not yet stored)
+
+A custom `DocumentChunker` can provide other keys and values in `metadata`. They will 
+be stored with the chunk and are accessible when it is returned in a search result or citation, within 
+`SearchResult.chunk_meta` / `Citation.chunk_meta`.
 
 ## Embed
 
--- a/docs/custom-pipelines.md
+++ b/docs/custom-pipelines.md
@@ -117,9 +117,7 @@ Chunks are returned with:
 - `embedding` - `None` (not yet embedded)
 - `document_id` - `None` (not yet stored)
 
-A custom `DocumentChunker` can provide other keys and values in `metadata`. They will 
-be stored with the chunk and are accessible when it is returned in a search result or citation, within 
-`SearchResult.chunk_meta` / `Citation.chunk_meta`.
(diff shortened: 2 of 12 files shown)


lawrenceakka:
The second commit here adds visualisation of all metadata to the inspector.  I have also written a (rather lengthy) test, which is probably unnecessary.

A further place in which metadata could be displayed is in the `haiku-rag search` command.

One issue with both of these is the duplication we have talked about before.  Chunk page numbers etc will appear in two places, unless we remove 'known' metadata before display. But this risks loss of user data if the scope of what is 'known' expands in future

haiku_rag_slim/haiku/rag/inspector/widgets/detail_view.py
```diff
@@ -93,6 +93,10 @@ async def show_chunk(self, chunk: Chunk) -> None:
         chunk_meta = chunk.get_chunk_metadata()
         content_parts.extend(self._format_provenance(chunk_meta))
 
+        if chunk.metadata:
+            metadata_str = "\n".join(f"  - {k}: {v}" for k, v in chunk.metadata.items())
+            content_parts.append(f"**Metadata:**\n{metadata_str}")
+
```
ggozad, line 111: This renders the standard provenance twice. `_format_provenance()` already displays `page_numbers`, `headings`, `labels`, and `doc_item_refs`, while both shipped chunkers store those same fields in `Chunk.metadata`.

It also prints the complete `doc_item_refs` list, bypassing the deliberate five-item truncation in `_format_provenance()`. Could we filter `ChunkMetadata.model_fields` from the raw metadata for display purposes? `chunk_meta` should remain lossless in the data model. The inspector does not need to display duplicate standard fields.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### Changed
```
ggozad, line 12: Let's put these changes under `### Added`

docs/python.md
ggozad: Could we update the search-result documentation here as well? This should mention that each result also carries the chunk’s verbatim stored metadata in `result.chunk_meta`, which is not shown to the model during QA.

ggozad:
The issue discussion explicitly requested coverage of the MCP serialization path. This test currently verifies only that the Python return values are `SearchResult` instances.

Please add a regression test that stores distinctive custom chunk metadata, invokes `search_documents` through the MCP serialization path, and asserts that the serialized result contains the unchanged `chunk_meta` value.

docs/custom-pipelines.md
```diff
@@ -115,6 +117,10 @@ Chunks are returned with:
 - `embedding` - `None` (not yet embedded)
 - `document_id` - `None` (not yet stored)
 
+A custom `DocumentChunker` can provide other keys and values in `metadata`. They will
```
ggozad: these two lines have trailing whitespace, let's keep paragraphs on a single line like the rest of the .md files

haiku_rag_slim/haiku/rag/store/models/chunk.py
```diff
@@ -132,12 +132,17 @@ class SearchResult(BaseModel):
 
     ``document_meta`` carries the parent document's metadata for citation
     consumers (UIs). Never part of ``format_for_agent`` output.
+
+    ``chunk_meta``is the anchor chunk's unparsed ``Chunk.metadata`` and does not
```
ggozad: insert space after ``chunk_meta``

ggozad (changes requested):

ggozad:
Thanks again @lawrenceakka, some minor things (the MCP test being the most important to help us remember to check if we will need a flag there in the future), otherwise great work :)

lawrenceakka:
> Please add a regression test that stores distinctive custom chunk metadata, invokes `search_documents` through the MCP serialization path, and asserts that the serialized result contains the unchanged `chunk_meta` value.

Sorry - I misunderstood what you said earlier.  Now done, along with the other fixes

ggozad:
Thanks, this addresses everything.

The `test` job is red, but it isn't your bug. `test_search_documents_preserves_chunk_meta_through_serialization` steps on a latent landmine in our test setup:

- `tests/test_cli.py` drives the Typer CLI, which calls the real `telemetry.configure()` (`cli.py:119`).
  That installs logfire's `WarnOnExtractTraceContextPropagator` as the process-global OTel textmap propagator.
- The in-process FastMCP `Client` propagates a trace context into the server.
- logfire warns `RuntimeWarning` on the first such extraction per process.
- Our `filterwarnings = ["error", "ignore::UserWarning", "ignore::DeprecationWarning"]` in `pyproject.toml`
  turns that into an exception inside the tool. FastMCP returns it as an error result and the client
  raises `ToolError`.

So it only fires when an xdist worker happens to run a CLI test before your MCP test. It passes in
isolation, and on a machine with enough cores it passes in the full suite too.

Add the following line on the test to fix

```python
@pytest.mark.filterwarnings("ignore:Found propagated trace context:RuntimeWarning")
```

PR #578, opened 2026-08-21
Use multiple databases simultaneously

ggozad:
Search, ask, analyze and chat across several LanceDB databases at once, instead of one per client.

## Configuring

```yaml
lancedb:
  databases:
    papers: s3://bucket/papers.lancedb
    wiki: s3://bucket/wiki.lancedb
    notes: /data/notes.lancedb
```

A location is a URI or a local path. `databases` and `uri` are mutually exclusive, and a blank name or location is rejected. `lancedb.uri` places a single unnamed database and takes the same kinds of location: a value with no scheme is a local path there too, so `haiku-rag init` creates it and every command that opens an existing database requires the path to exist. Results, citations, model input and errors opening a named database carry the configured name rather than the location, so a path or a bucket does not reach a trace or a model. Commands that report on a database — `info`, `init`, `tag` — print its location, as does an error about a path.

Databases searched together must have been written with the same embedder; two that disagree raise `ConfigMismatchError` naming both. A different `vector_dim` raises on open. A different provider or model name at the same dimension warns on a read-only open and raises on a writable one, since the same model served by another stack is spelled differently.

## Using

```python
results = await client.search("query")                      # every database
results = await client.search("query", sources=["papers"])   # one of them

answer, citations = await client.ask("...", sources=["papers", "wiki"])
result = await client.analyze("...", sources=["papers"])
```

`sources=None` covers the set, `sources=[]` covers nothing.

On the CLI, `search`, `ask`, `analyze` and `chat` cover the set. `settings`, `init-config` and `download-models` open no database. Every other command works on one, named with the global `--db-name NAME` — which resolves through `lancedb.databases`, so it reaches a database behind a URI that `--db PATH` cannot. `--db PATH` overrides the configured location for one invocation.

## Operations that need one database

Writing, rebuilding, vacuuming and visual grounding name a database: asking a set-covering client raises `AmbiguousDatabaseError`, and `create=True` is refused for the same reason. `clients_for(["name"])` returns a client for one of them, writable when the covering client is.

Converting, chunking and title generation are functions of the configuration rather than of a database, and work on a set — as does the embedder, which a set-covering client builds on first use and closes on teardown.

## How a search runs

Fetch, fuse, truncate, then enrich. The query is embedded once for the whole selection. The selected databases are searched concurrently, the union is fused into one ranked list — by the configured reranker, or by reciprocal rank fusion when none is configured — truncated to the requested limit, and only the survivors are enriched and context-expanded, each through the database it came from. Enrichment therefore costs what a single-database search costs rather than multiplying by the number of databases.

An image query is vector-only whatever type is requested. Full-text search embeds nothing, so it needs no agreement on embedders.

Cold databases open concurrently, so a first query costs the slowest open rather than their sum. A database that cannot be opened fails the whole query and is named in the error: a result set silently missing one of the databases asked for cannot be told apart from a complete one. Fused scores tie often, so equal scores keep the order they were fused in.

**Configure a reranker.** Reciprocal rank fusion compares ranks, not scores, so a database holding nothing relevant still contributes its best matches and displaces better ones. Measured on one corpus split three ways: retrieval MAP 0.9914 with a reranker against 0.9918 for the same corpus in a single database, and 0.6044 without one against 0.9798.

## Identity

`SearchResult.source`, `Citation.source` and `Document.source` name the database. `format_for_agent` emits a `Collection:` line, so the model can attribute evidence while it answers rather than only afterwards through the citations. Citation resolution routes a chunk id back to the database holding it, and the fallback for an id absent from the run's results looks only in the selected databases — a question scoped to some databases cannot cite another.

A chunk id is unique within a database and says nothing across them, so a database copied from another holds the same ids. In memory a result is told apart by the database and the id together. A citation cannot be: it records the id alone, so an id held by two of the databases searched raises `AmbiguousCitationError`, which the capabilities turn into a retry asking for other evidence. The direct-id fallback asks every database the question covers rather than taking the first that answers.

The analysis sandbox mounts the selected databases under the existing flat `/documents/{id}/` layout, resolving each id to its owning client. Document ids are UUID4, so the namespace stays unique; an id held twice is refused rather than resolved to whichever arrived last.

Capability instructions gain a short note about `source` only where the capability covers more than one database, so a single database is instructed byte for byte as before and no existing run changes.

The chat document filter selects by document and database, labelling each with its name. It previously matched the displayed title or URI as a substring, which widened a selection to documents the user did not pick, across every database. Applying a selection narrows the question to the databases it names and filters by id within them; a twin id inside another selected database still matches there, since a serialized id filter cannot carry a source. Independently built databases use UUID document ids and do not collide.

## Structure

`DatabaseScope` resolves which databases an operation covers, once, from the configuration plus at most one selector, and never rewrites the configuration to say it. `HaikuRAG` stays the public surface over one of two sessions: `SingleDatabaseSession`, which owns a store and its repositories and is the only thing that writes, or `FederatedSession`, which composes single sessions and reads. The write implementations are typed against the single session, so a federated write is unrepresentable rather than guarded against at each call.

## Deliberately deferred

- **Cross-database deduplication.** A document present in two databases appears twice.
- **Copied databases.** Chunk-id and document-id collisions are refused where a wrong answer would be silent — citations and the analysis mount.
- **Over-fetch scaling.** Each database over-fetches the same multiple, so a reranker scores candidates in proportion to the number of databases. Measured on ORB split three ways with a multimodal reranker: 2x a single database, and 5.5x with no reranker. Fine at the handful of databases this targets, and a knob to revisit before large fan-out.
- **One embedder per process.** `Store` builds its own, so N databases mean N embedders, and N weight loads for a local `sentence-transformers` model. Sharing needs reference-counted lifecycle, since embedders own closeable clients.
- **Authorization.** haiku.rag searches the databases it is handed. Who may see which stays with the caller.

## Verification

2603 tests at 100% coverage, plus 21 integration and 269 evaluation tests. Exercised end to end against two evaluation databases sharing an embedder (OpenRAG Bench, 113,881 chunks; HotpotQA, 70,527 chunks): federated FTS and hybrid search, `sources` selection, `ask` with citations labelled by database, and the analysis sandbox mounting 67,581 documents from both.

Regression arms on ORB, 3,045 queries, comparing one database against the same corpus sharded three ways: sharded with a reranker matches single-database retrieval (ΔMAP +0.0004), and QA is within noise at n=250.


lawrenceakka:
I have been keeping an eye on this!  It is impossible at the moment to have the default DB be anything other than `haiku.rag.lancedb` (though of course the data store *directory* is configurable). I am wondering whether as part of this, you could make the actual DB *name* configurable, even if there is only one db. Strictly, this patch covers rather different functionality, though it does seem to me to be related.  If you prefer me to open a different issue (or PR instead, once this one has been merged) let me know. And if you prefer me to drop the idea completely, let me know that as well!

ggozad:
> I have been keeping an eye on this! It is impossible at the moment to have the default DB be anything other than `haiku.rag.lancedb` (though of course the data store _directory_ is configurable). I am wondering whether as part of this, you could make the actual DB _name_ configurable, even if there is only one db. Strictly, this patch covers rather different functionality, though it does seem to me to be related. If you prefer me to open a different issue (or PR instead, once this one has been merged) let me know. And if you prefer me to drop the idea completely, let me know that as well!

You can already name it whatever you like on main, in two ways.

Per invocation, `--db` works on every command:

```bash
haiku-rag add-src report.pdf --db /data/my.lancedb
```

As a configured default, `lancedb.uri` takes a plain local path, not only a `db://` or `s3://` URI:

```yaml
lancedb:
  uri: /data/my.lancedb
```

With that, every command uses `/data/my.lancedb` and no flag is needed. In Python it is `HaikuRAG("/data/my.lancedb")`.

`storage.data_dir / haiku.rag.lancedb` is only the fallback when neither is set. The real gap is the docs, which present `lancedb.uri` under remote and object storage only. Opened #582 for that. If `uri` misses something your case needs, tell me what and I will take it as an issue.

lawrenceakka:
This is great, thanks.   Sorry I missed it

PR #590, opened 2026-08-31
Honor configured vector metric during search

### LINKED ISSUE #588: Configured vector distance metric is not honored consistently in search
## Summary

`search.vector_index_metric` is used when the IVF-PQ index is created, but `ChunkRepository.search()` does not explicitly set the same distance type on the vector query.

With the currently pinned LanceDB 0.34.0, an unindexed vector query defaults to L2. Once an index exists, LanceDB can use the index metric instead. As a result, retrieval semantics can depend on whether the table has an index, even though the application configuration has not changed.

There is a related score-conversion problem for `dot`: `_process_search_results()` applies `1 / (1 + distance)` to every vector metric. Dot distances can be negative, so this conversion is not guaranteed to remain monotonic and may give a better result a lower score.

## Relevant code

The configured metric is applied while creating the index:

```python
config=IvfPq(
    distance_type=self._config.search.vector_index_metric,
)
```

But the search query currently does not set it:

```python
results = (
    self.store.chunks_table.query()
    .nearest_to(query_embedding)
    .column("vector")
    .refine_factor(self.store._config.search.vector_refine_factor)
)
```

And all vector distances use one conversion:

```python
((df["_distance"] + 1).rdiv(1)).clip(lower=0.0)
```

## Minimal reproduction

Using LanceDB 0.34.0 without a vector index:

```python
import lancedb
from tempfile import TemporaryDirectory

with TemporaryDirectory() as path:
    db = lancedb.connect(path)
    table = db.create_table(
        "chunks",
        data=[
            {"id": "cosine-best", "vector": [10.0, 0.0]},
            {"id": "l2-best", "vector": [1.0, 1.0]},
        ],
    )
    query = [1.0, 0.0]

    default = table.search(query).limit(2).to_list()
    cosine = table.search(query).distance_type("cosine").limit(2).to_list()

    print([row["id"] for row in default])
    print([row["id"] for row in cosine])
```

Observed output:

```text
['l2-best', 'cosine-best']
['cosine-best', 'l2-best']
```

The first result follows L2, despite `haiku.rag` defaulting `vector_index_metric` to `cosine`.

For dot distance, LanceDB returns `1 - dot_product`. For example, a dot product of 10 produces a distance of -9. The current conversion maps that to a negative value and then clips it to 0, while a lower dot product of 1 produces distance 0 and score 1. This reverses the score relationship. Since scores are later used by context expansion ordering and picture-result deduplication, this can affect more than display formatting.

## Expected behavior

- Vector queries explicitly use `config.search.vector_index_metric` for both vector and hybrid retrieval.
- Flat and indexed search use the same metric and produce consistent Top-K ordering for the same data.
- Score conversion remains monotonic for `cosine`, `l2`, and `dot`.
- Changing the configured metric should clearly require rebuilding an existing vector index.

A possible query-side change would be:

```python
.distance_type(self.store._config.search.vector_index_metric)
```

The score representation may need a metric-aware conversion, or a separation between the native value used for ordering and the user-facing relevance score.

## Suggested regression tests

1. Parameterized flat-search ranking for `cosine`, `l2`, and `dot` using controlled, non-normalized vectors.
2. Top-K agreement before and after building an index with the configured metric.
3. Monotonic score assertions for every supported metric, especially negative dot distances.
4. Coverage of the vector branch used by hybrid retrieval.

## Environment

- haiku.rag `main`: `b87ae16910035144ad64908219962b59e115d228`
- LanceDB: `0.34.0`
- Python: `3.12.3`

I would be happy to submit a PR with the query fix and regression tests once the desired score semantics and scope are confirmed.

icearia0219:
## Summary

- Apply search.vector_index_metric to every LanceDB vector query, keeping flat and indexed retrieval on the configured metric.
- Remove the unused dot option from configuration and current documentation.
- Add regression coverage with non-normalized vectors whose cosine and L2 rankings differ, then verify the same Top-K before and after building an IVF-PQ index.

## Why

The vector index was built with search.vector_index_metric, but ChunkRepository.search() did not pass that metric to nearest_to(). LanceDB therefore used its query-side default for flat search. Unit-normalized embeddings masked this because cosine and L2 produce the same ranking for those vectors, while non-normalized vectors expose the mismatch.

Per the discussion in #588, metric changes against an existing incompatible index are left to LanceDB: it warns and falls back to brute-force search with the requested metric. Rebuilding the index restores indexed performance.

## Validation

- PYTHONUTF8=1 uv run --locked pytest tests/test_chunk.py tests/test_config.py -q -k "not test_find_config_file_env_var_tilde_expansion" — 175 passed
- uv run --locked ruff check — passed
- uv run --locked ruff format --check on the changed Python files — passed
- uv run --locked ty check on the changed Python files — passed

The excluded config test changes HOME to test tilde expansion, which does not control Path.home() on Windows. A broader Windows run also reaches a pre-existing POSIX-only filename test (a question mark in a SQLite filename); Linux CI remains the authoritative full-suite run.

Closes #588

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 
 - Migration to 0.38.0 no longer fails with `UnicodeDecodeError` on a
   `docling_document` blob written as zstd.
+- Vector search explicitly uses `search.vector_index_metric`, so flat and indexed
+  searches rank with the same configured distance metric. The unused `dot` option
+  is removed; `cosine` and `l2` remain.
 
 ## [0.79.0] - 2026-08-28
 
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -391,7 +391,7 @@ haiku-rag create-index [--db /path/to/your.lancedb]
 
 **Requirements:**
 - Minimum 256 chunks required for index creation (LanceDB training data requirement)
-- Creates an IVF_PQ index using the configured `search.vector_index_metric` (cosine/l2/dot)
+- Creates an IVF_PQ index using the configured `search.vector_index_metric` (cosine/l2)
 
 **When to use:**
 - After ingesting documents (indexes are not created automatically)
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -114,7 +114,7 @@ qa:
 search:
   limit: 5                     # Default number of results to return
   max_context_chars: 5000     # Maximum characters in expanded context
-  vector_index_metric: cosine  # cosine, l2, or dot
+  vector_index_metric: cosine  # cosine or l2
   vector_refine_factor: 30
 
 doctor:
--- a/docs/configuration/storage.md
+++ b/docs/configuration/storage.md
@@ -300,7 +300,7 @@ Configure vector search settings:
 
 ```yaml
 search:
-  vector_index_metric: cosine  # cosine, l2, or dot
+  vector_index_metric: cosine  # cosine or l2
   vector_refine_factor: 30     # Re-ranking factor for accuracy
 ```
 
@@ -309,7 +309,6 @@ For search behavior settings (`limit`, `max_context_chars`), see [Search and Que
 - **vector_index_metric**: Distance metric for vector similarity:
   - `cosine`: Cosine similarity (default, best for most embeddings)
   - `l2`: Euclidean distance
-  - `dot`: Dot product similarity
 - **vector_refine_factor**: Improves accuracy when using a vector index by retrieving `refine_factor * limit` candidates (using approximate search) and re-ranking them with exact distances. Higher values increase accuracy but slow down queries. Default: 30
   - **Only applies with a vector index** - has no effect on brute-force search, which already returns exact results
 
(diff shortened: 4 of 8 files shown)


CHANGELOG.md
```diff
@@ -2,15 +2,13 @@
...
-### Added
-
-- `--full-citations` on `haiku-rag ask` and `haiku-rag analyze` renders citation
-  text untruncated. `format_citations_rich` takes a `full` argument.
-
 ### Fixed
 
 - Migration to 0.38.0 no longer fails with `UnicodeDecodeError` on a
   `docling_document` blob written as zstd.
+- Vector search explicitly uses `search.vector_index_metric`, so flat and indexed
```
ggozad, line 9: Very minor but could you please amend your commit and change the changelog to:

```
- Vector search sets `search.vector_index_metric` on every query.
- `dot` removed from `search.vector_index_metric`; switch to `cosine` or `l2` and rerun `create-index`.
```

ggozad (changes requested):

ggozad:
Thank you @icearia0219 this looks great! I left one minor comment, please amend and I will merge asap, it will go to next release.

ggozad:
Never mind, I will do it so you don't need to rebase.

PR #596, opened 2026-09-01
Fix for missing timezone info on ingested docs. mcp schema requires TZ

lawrenceakka:
When a document is saved, haiku.rag stamps it with a "created at" and "updated at" time using `datetime.now()`. That gives a time with no timezone attached, for example 2026-09-01T16:56:29.958257.

The `get_document` MCP tool publishes a JSON schema that says these two fields must be a date-time, which in context means RFC 3339, not plain ISO 8601. RFC 3339 requires the timezone to be stated, either as Z for UTC or as an offset like +01:00; ISO 8601 permits omitting it. This does not seem to be enforced by a lot of tools, including `FastMCP`'s client.

Claude Code's MCP client does enforce it however, and calling `get_document` therefore errors with:

>Error: Structured content does not match the tool's output schema: data/result/created_at must match format "date-time", data/result/updated_at must match format "date-time", data/result must be null, data/result must match a schema in anyOf ...

I have added a `field_validator` on `Document.created_at`/`updated_at` that attaches UTC to any naive value at construction time, and switched every write site to `datetime.now(UTC)`. 

I think there is a risk for documents which have been ingested before this fix. Whilst we can treat a date without a TZ as UTC and correct for it, that of course assumes that the server's local time is indeed UTC.  The displayed time on old records from a non-UTC database will therefore be out by a fixed number of hours.

### DIFF
--- a/haiku_rag_slim/haiku/rag/store/models/document.py
+++ b/haiku_rag_slim/haiku/rag/store/models/document.py
@@ -1,8 +1,8 @@
 import json
-from datetime import datetime
+from datetime import UTC, datetime
 from typing import TYPE_CHECKING
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, field_validator
 
 from haiku.rag.store.compression import compress_docling_split, decompress_json
 
@@ -29,8 +29,19 @@ class Document(BaseModel):
     docling_document: bytes | None = Field(default=None, exclude=True)
     docling_pages: bytes | None = Field(default=None, exclude=True)
     docling_version: str | None = Field(default=None, exclude=True)
-    created_at: datetime = Field(default_factory=datetime.now)
-    updated_at: datetime = Field(default_factory=datetime.now)
+    created_at: datetime = Field(default_factory=lambda: datetime.now(UTC))
+    updated_at: datetime = Field(default_factory=lambda: datetime.now(UTC))
+
+    @field_validator("created_at", "updated_at")
+    @classmethod
+    def _assume_utc(cls, value: datetime) -> datetime:
+        """Attach UTC to a timezone-less timestamp instead of leaving it ambiguous.
+
+        A row written before timestamps carried an explicit offset parses back
+        with no tzinfo. Treat it as UTC rather than pass it on naive, which the
+        MCP tools' declared JSON schema rejects as an invalid date-time.
+        """
+        return value if value.tzinfo else value.replace(tzinfo=UTC)
 
     def set_docling(self, docling_doc: "DoclingDocument") -> None:
         """Serialize and store a DoclingDocument, splitting structure and pages.
--- a/haiku_rag_slim/haiku/rag/store/models/document.py
+++ b/haiku_rag_slim/haiku/rag/store/models/document.py
@@ -34,14 +34,8 @@ class Document(BaseModel):
 
     @field_validator("created_at", "updated_at")
     @classmethod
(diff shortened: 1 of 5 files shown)


haiku_rag_slim/haiku/rag/store/models/document.py
```diff
@@ -29,8 +29,19 @@ class Document(BaseModel):
...
+    @field_validator("created_at", "updated_at")
+    @classmethod
+    def _assume_utc(cls, value: datetime) -> datetime:
+        """Attach UTC to a timezone-less timestamp instead of leaving it ambiguous.
+
+        A row written before timestamps carried an explicit offset parses back
+        with no tzinfo. Treat it as UTC rather than pass it on naive, which the
+        MCP tools' declared JSON schema rejects as an invalid date-time.
+        """
+        return value if value.tzinfo else value.replace(tzinfo=UTC)
```
ggozad: `replace(tzinfo=UTC)` asserts that a legacy timestamp was already UTC. Those values came from datetime.now() and represented host-local wall time.

On a UTC+3 host, a legacy 10:36 becomes 10:36Z, while a subsequent update may be written as 07:40Z, producing `created_at > updated_at`. `update_meta()` then persists that incorrect interpretation.

This preserves the intent I think:
`return value if value.tzinfo else value.astimezone(UTC)`

haiku_rag_slim/haiku/rag/store/models/document.py
```diff
@@ -29,8 +29,19 @@ class Document(BaseModel):
...
     docling_version: str | None = Field(default=None, exclude=True)
-    created_at: datetime = Field(default_factory=datetime.now)
-    updated_at: datetime = Field(default_factory=datetime.now)
+    created_at: datetime = Field(default_factory=lambda: datetime.now(UTC))
+    updated_at: datetime = Field(default_factory=lambda: datetime.now(UTC))
+
+    @field_validator("created_at", "updated_at")
+    @classmethod
+    def _assume_utc(cls, value: datetime) -> datetime:
+        """Attach UTC to a timezone-less timestamp instead of leaving it ambiguous.
```
ggozad: I try to avoid comments that narrate the history of why a change was made. I am not particularly good at it and AI assistants love these.
The validator is clear enough without the docstring, so please remove or make it a one line describing the invariant.

ggozad (changes requested):
Thanks for the clear write-up. Found some issues that need to be addressed.
Like before, awesome work :)

1.  Add regression coverage for both failure modes:

The naive-timestamp branch is not asserted. Please add two unit assertions:

- Under a fixed non-UTC TZ, assert the exact UTC conversion of a fixed naive timestamp. This distinguishes `astimezone(UTC)` from the incorrect `replace(tzinfo=UTC)`.

- Assert that serialization retains timezone information:

```
value = doc.model_dump(mode="json")["created_at"]
assert datetime.fromisoformat(value).utcoffset() is not None
```
The second assertion covers the reported missing-offset defect without depending on FastMCP/jsonschema format checking, which is inactive in the current dependency stack.

Ensure the test restores TZ and calls `time.tzset()` afterward so libc timezone state does not leak within the xdist worker.

2. Add an entry to the CHANGELOG
Something like 
```
### Fixed

  - Document timestamps include timezone information and satisfy the MCP
    date-time output schema.
```

ggozad:
@lawrenceakka I am so sorry, I was buried with 0.82.0 and forgot to merge yours too. 
Will do asap

ggozad (approved):

Issue #599, opened 2026-09-03
Some thoughts about MCP - for discussions


lawrenceakka:
Hi @ggozad

I have been thinking about the MCP tools. I have jotted down a list of issues and potential changes, not in any particular order, which may be worth thinking about/discussing.  Let me know what you think, when you have time.

1. The docstrings for the current tools are quite brief. Together with the parameter schema, they are exposed to the MCP client by FastMCP as the client's only guidance on what a tool does and when to use it. In my experience with Claude, it usually works out the right tool to use from this alone, but not always. I think the docstrings would benefit from being longer perhaps with an explicit "use this to do X" sentence, not just a description of the parameters.

2. Relatedly, the purpose of the MCP server as a whole is not exposed to the client anywhere. The MCP spec provides for a server-level `instructions` field, which FastMCP can set. I have seen conflicting reports on whether Claude actually surfaces this to the model: https://github.com/anthropics/claude-code/issues/23808 A workaround, if it is genuinely unreliable, is to put more of that context into each tool's own description instead. Not sure this is worth the duplication it causes across tools.

3. Following on from that, I find I have to remind Claude each session to actually use the knowledge base at all before it will reach for these tools unprompted. It would help if a tool's own description said something like "use this whenever asked to find, look up, or check something that might be in the document store" (needs refining), so that at least something to that effect is guaranteed to be in context from the tool listing itself, rather than relying on the user to say it every time. This could go in the instructions field, if it were actually to work reliably.

4. A couple of tools expose a `filter` argument, described only as "Optional SQL WHERE clause to filter documents." They do not say which columns are actually available to filter on (e.g. `uri`, `title` for documents; `document_id`, `content`, `order` for chunks). It would help at least to state the real column list in the docstring, and since that list is static per table, it could plausibly be generated automatically from the Arrow schema at server start rather than kept in sync by hand.

5. `search_documents` has no `filter` parameter at all, unlike `list_documents` and others. It would be useful to be able to scope this to one document or a subset of documents.

[...]

PR #600, opened 2026-09-04
Make the MCP server read-only and multi-database; replace ask_question and analyze with execute_code; ship a Claude Code and Codex plugin with the haiku-rag skill

### LINKED ISSUE #604: Expose the analysis sandbox as an MCP tool
`ask_question` and `analyze` run an agent inside the MCP server. The server needs its own model configuration and its own credentials, and it answers with prose the client then has to trust. In an agentic client the client already is the model, and the skill teaches it the search-and-cite loop, so both tools are a second model doing worse what the first one is there to do.

What the server can do that no client can is run code next to the data. `Sandbox` mounts every document in scope as a read-only filesystem and executes Python against it in a subprocess interpreter. That is the server's actual added value, and it is not exposed. `Sandbox(db_path, config, context)` is already a standalone API, with no capability, agent or model attached.

ggozad:
Refs #599. Closes #604.

## Summary

- The MCP server becomes read-only: `add_document_from_file`, `add_document_from_url`, `add_document_from_text` and `delete_document` are removed. Ingestion is `haiku-rag add` / `add-src` / `delete` and `haiku-ingester`.
- `haiku-rag mcp` covers the configured `lancedb.databases` set, as `search`, `ask` and `analyze` already do. Read tools take `sources` / `source`; results and listings carry the collection name.
- Failures are MCP errors (`isError: true`) carrying the error's message instead of empty values: `No document with id …`, `No section …`, `Invalid base64 image`, an unknown collection, a filter the query engine rejects, a failing `execute_code` program with the error it hit and what it printed.
- `search_documents` expands results to their section (`HaikuRAG.expand_context`, as every other consumer of search results already did) and returns the rendering the in-process agents get (`format_for_agent`: rank, title, headings, the matched chunk's metadata, passage) plus `ImageContent` blocks for pictures, and no structured content: Claude Code and the Agent SDK drop text blocks when `structuredContent` is present, Desktop shows both. Every result's text carries `Document ID` and, over several databases, `Collection`.
- Two navigation tools built on `document_items`: `get_document_outline` (heading tree with page numbers) and `get_document_section` (one section's text, subsections included). No chunk ids leave the server.
- Server `instructions`, `version`, `ToolAnnotations` and full parameter descriptions on every tool. `filter` on both search tools. `list_documents` returns each document's `metadata`.
- `ask_question` and `analyze` are replaced by `execute_code(code, filter, sources)`: one Python program per call in the analysis sandbox over the selected documents, returning what it printed. No model runs on the server; in Claude Code the client is the model, and the skill carries the workflow. The sandbox, shared with the analysis capability, gains `chunk_meta` on in-code `search()` results, `metadata` on `list_documents()` rows and in `metadata.json`, and a per-document `chunks.jsonl` (`chunk_id`, `metadata`), which is how chunk metadata becomes filterable, from code.
- A Claude Code and Codex plugin under `plugins/haiku-rag/`: one `.mcp.json` and one `haiku-rag` Agent Skill, with a manifest per client (`.claude-plugin/plugin.json`, `.codex-plugin/plugin.json`) and a marketplace per client (`.claude-plugin/marketplace.json`, `.agents/plugins/marketplace.json`). Install with `claude plugin marketplace add ggozad/haiku.rag` then `claude plugin install haiku-rag`, or `codex plugin marketplace add ggozad/haiku.rag` then `codex plugin add haiku-rag@haiku-rag`. Both manifests carry the package version; `scripts/bump_version.py` bumps them.
- `fastmcp>=4.0.2,<5.0.0` on MCP Python SDK 2. The server answers both the session-based and the sessionless (2026-07-28) protocol.
- `pydantic-monty>=0.0.23`: the analysis sandbox gains `collections`, `itertools`, `functools`, `dataclasses`, function decorators and `str.format`. Monty 0.0.22+ caps host callbacks per checkout at 1000 by default; `Sandbox._session_limits` sets `max_suspensions` out of reach, since `analysis.code_timeout` governs. The `execute_code` description, the skill, the capability instructions and `docs/mcp.md` state the same interpreter limits.

## Tool set

| Tool | Registered | Parameters |
|---|---|---|
| `search_documents` | always | query, limit, filter, include_images, sources |
| `search_documents_by_image` | multimodal embedder only | image_base64, limit, filter, include_images, sources |
| `get_document` | always | document_id, source |
| `get_document_outline` | always | document_id, source |
| `get_document_section` | always | document_id, section_id, source |
| `list_documents` | always | limit, offset, filter |
| `execute_code` | always | code, filter, sources |

## Issue #599, point by point

| # | Point | Resolution |
|---|---|---|
| 1 | Docstrings too brief | Every tool: what it does, when to pick it over its siblings, Google `Args:` so each parameter carries a description in the schema. |
| 2 | Server-level `instructions` | Set. Claude Code surfaces them (1.0.52), Claude Desktop does not (anthropics/claude-code#23808), so every description stands alone. |
| 3 | Proactive use of the knowledge base | Instructions and the plugin's skill both say to check the knowledge base before answering from memory. |
| 4 | Which columns a `filter` can use | Listed in every `filter` description, derived from `DocumentMetaRecord`: `id, uri, title, metadata, created_at, updated_at`. `metadata` is a JSON string, matched with LIKE. |
| 5 | `filter` on `search_documents` | Added, forwarded to `HaikuRAG.search(filter=)`. |
| 6 | Filter by keys inside chunk metadata | Out of scope. `filter` is evaluated against `document_meta` and rewritten to `document_id IN (...)`; chunk metadata is not reachable by any filter today, and there is no scalar `page_no` (chunks carry `page_numbers: list[int]`). Making it filterable is a storage change (Lance JSON type or a LabelList tag column, migration, query plumbing) with its own evaluation. Reachable from code instead: `execute_code` reads `chunks.jsonl` and filters in Python. |
| 7 | Read a document in order, with metadata | `get_document` already returns the document in reading order (docling markdown). `get_document_outline` / `get_document_section` add navigation by document structure; chunks stay internal. |
| 8 | Score scale undocumented | Descriptions state that scores are not comparable across queries; the text rendering shows rank, not score. |
| 9 | Hide `ask_question` / `analyze` | Both are removed; `execute_code` needs no model. (`qa.model` always has a default, so "no model configured" could not have been detected from config.) |

## Breaking changes

- MCP write tools removed. Use `haiku-rag add`, `haiku-rag add-src`, `haiku-rag delete`, or `haiku-ingester`.
- `create_mcp_server(db_path=None, config=None)`: `read_only` removed, the client always opens read-only.
- MCP tools `ask_question` and `analyze` removed; `execute_code` added. `haiku-rag ask` / `analyze` and the capabilities are unchanged.
- `format_citations` removed from `haiku.rag.utils` (its only caller was `ask_question`); `format_citations_rich` stays.
- `search_documents` / `search_documents_by_image`: text content is the agent rendering, pictures arrive as `ImageContent` blocks, no structured content.
- Tools raise on failure, with the error's message; an empty result no longer doubles as an error.
- `haiku-rag mcp` over several configured databases no longer refuses to start.

## Out of scope

- Chunk-level metadata filtering as a query (point 6), see above.
- MCP resources, prompts, progress reporting.

## Commits (each self-contained)

1. Remove write tools; server opens read-only.
2. Cover the configured database set; `sources` / `source`; collection on citations.
3. Descriptions, instructions, version, annotations; `filter` on search; strict base64; `DocumentInfo.metadata`.
4. Error contract.
5. `get_document_outline` / `get_document_section`; `ask_question` always cites.
6. Search result shape: text + pictures + structured; `Document ID` in every result.
7. `--no-agents`.
8. Claude Code plugin and skill.
9. fastmcp 4 and MCP SDK 2.
10. Search results expanded; matched chunk metadata rendered.
11. Search results text-only; filter error lists our columns; untitled citations name the URI once.
12. `execute_code` in place of `ask_question` / `analyze`; `--no-agents` goes; the sandbox exposes chunk metadata.
13. `toc.json` `item_range` is the documented line slice into `items.jsonl`; docs name the tools that take `sources`.
14. Codex plugin from the same bundle: `plugins/haiku-rag/` with both manifests, `.agents/plugins/marketplace.json`, Codex docs.
15. MCP guidance aligned with the analysis instructions; Monty 0.0.23 with the host-call cap raised.
16. One sandbox contract: host errors keep their message for every caller; `analysis.code_timeout` checked before every host call, not only database-bound reads.
17. One error contract for the server: no masking, no filter pre-check, no error translation; every failure carries its message.

## Verification

- `tests/test_mcp.py` through `fastmcp.Client` (wire serialization, `initialize_result.instructions`, annotations, parameter descriptions, `is_error`), including a two-database fixture and a search → outline → section composition test driven from the search text alone.
- Full non-integration suite: 2750 passed (1 skipped), coverage 100%.
- `claude plugin validate --strict` passes on the plugin and the marketplace manifest; the Agent Skills reference validator passes on the skill. Codex: `codex plugin marketplace add` of the repo's local marketplace, `codex plugin add haiku-rag@haiku-rag`, a `$haiku-rag` call reaching `list_documents` through the plugin's server, temporary install removed.
- Smoke over a real stdio subprocess (`uv run haiku-rag mcp --stdio --db …` driven by `fastmcp.Client`): server name and version, instructions, tool list, `list_documents`, and an `isError` for an unknown document. For `execute_code`, against two local databases (34 documents): a program counting documents per collection, reading `chunks.jsonl` and `chunk_meta`; `sources` narrowing; a failing program returning its partial output; an unknown collection as an error.
- Exercised from Claude Code via `claude --plugin-dir` against two local databases: skill loaded, instructions surfaced, every tool driven from the rendered text, errors visible.


lawrenceakka:
> ## Summary
> * The MCP server becomes read-only: `add_document_from_file`, `add_document_from_url`, `add_document_from_text` and `delete_document` are removed. Ingestion is `haiku-rag add` / `add-src` / `delete` and `haiku-ingester`.

Comments on this at #599 - is it architecturally necessary to remove these?  I think they might be useful

> * `haiku-rag mcp` covers the configured `lancedb.databases` set, as `search`, `ask` and `analyze` already do. Read tools take `sources` / `source`; results, listings and citations carry the collection name.

> * `search_documents` returns the rendering the in-process agents get (`format_for_agent`: rank, title, headings, passage) plus `ImageContent` blocks for pictures, with the `SearchResult` list as structured content. Every result's text carries `Document ID` and, over several databases, `Collection`.

`format_for_agent` does *not* receive all the custom chunk metadata.  The LLM client should be able to see all the metadata, preferably in some sort of structure.  For me, that is what carries para and footnote numbers etc.  The LLM (Claude) is very good at including and interpreting this metadata in its citations (and saying, for example, that a particular fact is from paragraph XX of YY document). 

> * Two navigation tools built on `document_items`: `get_document_outline` (heading tree with page numbers) and `get_document_section` (one section's text, subsections included). No chunk ids leave the server.

Depends what we mean by "Section".  I have also found that docling sometimes is not very good at identifying sections.   Sometimes it classifies a normal para as a section, with rather odd results!  

> * Server `instructions`, `version`, `ToolAnnotations` and full parameter descriptions on every tool. `filter` on both search tools. `list_documents` returns each document's `metadata`.

Great - for document level metadata, but chunk level metadata is also useful

> * `haiku-rag mcp --no-agents` leaves `ask_question` and `analyze` unregistered.

This can be done via FastMCP [tags](https://gofastmcp.com/servers/tools#decorator-arguments)

> * A Claude Code plugin under `claude-plugin/` (server config + `haiku-rag` skill), installable with `claude plugin marketplace add ggozad/haiku.rag`.
>

ggozad:
Thanks @lawrenceakka, this is great feedback. Not quoting to make this readable again.

**Writes.** I am going to resist this one a bit longer: 

* LanceDB allows one writer per database URI, so a server that writes cannot sit next to `haiku-ingester`, and each Claude window is its own stdio server, so several clients writing is the accidental default. 
* Over streamable HTTP there is no authentication yet, and the old `add_document_from_file` / `add_document_from_url` dispatched any non-http string as a server-side path, so a remote client could ingest a file the server can read and then read it back. I had this back in my head as something to look at eventually and now that the opportunity arose, my thinking was that I would not trust an agent with this.

Having said that, your projects use case is real, but the part that makes it work is deciding who may write where, and that is not something the MCP server can decide on its own. Writing goes through the CLI or the ingester, which already know about one writer. If this turns out to be a blocker for you rather than a nice-to-have, an opt-in flag (or tags like you suggest) with `add_document_from_text` and `delete_document` is small.

**Expanding around a hit.** Agreed, and looking at it, the MCP server was the only consumer of search results that did not expand them: the agents, the analysis sandbox and the inspector all do. Search results will now always come expanded, the same section-bounded expansion the agents get (`HaikuRAG.expand_context`, budgeted by `search.max_context_chars`). I think this also settles the chunks question, what comes back is the passage in its section, not a chunk. I am not sure how much you have experimented with the non-MCP parts and seen how context expansion works, but it has proven a lot more reliable in providing good context than trying to go up and down iterating chunks.

**Chunk metadata.** You are right, and I should have caught it. `format_for_agent` deliberately omits `chunk_meta` for the in-process agents. I will render it in the MCP text, labelled as the matched chunk's metadata, the same opt-in way `Document ID` is rendered so the agents' rendering does not change. Labelled, because an expanded passage spans several chunks and only the one that matched carries its metadata through. If you need the metadata of every chunk under a passage, that is an items-to-chunks lookup after expansion and a separate change; tell me if the matched chunk's is not enough for your citations.

**Sections.** The outline is only as good as docling's heading detection, and flat or odd outlines do happen. With expanded results a client has three ways in: the whole document, the neighbourhood of a hit, and the outline. I do not want to add a fourth.

**Skill.** here's how it is useful:

* It is the Claude Code way. The plugin installs the server and the skill in one step, and the skill carries the workflow that spans tools: search, then outline, then section; when to reach for `ask_question` instead of reading. No single tool description can say that, and server instructions are not shown by every client.
* It loads on demand. Claude Code keeps only the skill's description in context and pulls in the body when a request matches it, or when the user types `/haiku-rag`. Tool descriptions and instructions sit in context on every turn, whether or not the knowledge base is relevant.
* Its description is the answer to your point 3. "Use whenever a request could be answered from the user's documents" is how Claude Code decides to check the knowledge base before answering from memory.
* It pre-approves the tools. They are all reads, so the user is not asked to allow each call.

The descriptions still stand alone, so a client without the skill is not worse off. A read-me-first tool spends a call per session and only works if the model decides to call it, it's how we used to do things a while ago, but that's the kind of thing skills replace.

**Tags.** I hadn't noticed tags, thanks :) The branch is on fastmcp 4 now, where `include_tags` / `exclude_tags` are gone in favour of `server.disable(tags=...)`. If writes come back I will tag both groups.

lawrenceakka:
Hi @ggozad,

I have tried this out, and it all works very well, even without the skill.  Thank you for looking at it so quickly.

A couple of small comments:

**Writes**  Yes, I did wonder if there might be an architectural issue. Writing arbitrary paths does not sound like a good idea!  I think, for me, concurrent writes will not be a problem, but I can see that they may be in the general case.  If it turns out that I need to upload via MCP, I can add write the tools quite simply in my own server.

**Expansion** Works great. It uses more context that necessary (for me) on the first tool call, but in fact often avoids later calls becuase the material is already in context.  In any case, as you say, the more semantic information the better in general.

**Chunk metadata**  Also working great. Claude is still able to provide pinpoint citations.  And no, I do not think we need metadata for every chunk in the expanded answer. 

As part of my "Campaign for More Metadata" I have noticed that it is not possible to filter on chunk metadata (though it is on document metadata).  How do you feel about adding something like a `chunk_filter` on `search_documents`/`list_documents` to allow for `metadata LIKE '%"para_no": "14"%'` ? To return a particular para, Claude has resorted to counting the paras one-by-one, which is not particularly reliable. I have not yet persuaded it to do this by searching for the number 14.

The downside of this, as previously discussed is that there is no index for this so there has to be a string scan for every chunk.  Also, I think one has to be careful about the query text since `'%"para_no": "14"%'` would also match `"para_no": "140"`  Since we do not have a means of getting all of a documents chunks (yet?) I cannot easily implement this by fetching a document's chunks by document_id and filtering in Python on the parsed metadata["para_no"]


**Skill**. FIne. As I say, the expanded docstrings seem to do the trick for me.  Obviously no harm in having a skill as well though.

**Visual grounding** I'm wondering about a tool exposing `HaikuRAG.visualize_chunk` for a given chunk.  Frankly, I'm not sure I'd use it, but the data is already there.  Would take up a fair bit of context though ....

ggozad:
Thanks for testing @lawrenceakka.

I think you will be pleasantly surprised when you actually try the skill, it will give you better results:

From a checkout of this branch, for one session: `claude --plugin-dir /path/to/haiku.rag/claude-plugin`. `haiku-rag` has to be on the PATH and your `haiku.rag.yaml` specifies the databases, as with your own registration.

Once merged: `claude plugin marketplace add ggozad/haiku.rag`, then `claude plugin install haiku-rag@haiku-rag`. But we are not there yet.

If you keep your own server registration instead of the plugin: copy `claude-plugin/skills/haiku-rag` into `~/.claude/skills/` and change the tool prefix in its `allowed-tools` from `mcp__plugin_haiku-rag_haiku-rag__` to `mcp__<your server name>__`.

Once you do that, `/haiku-rag` appears in the command list, the tools stop asking for permission, and when a request matches the skill's description ("could be answered from the user's documents") Claude loads the body and follows the workflow in it: search, then outline and section for the long documents, filters, how to cite. Without it, the tool descriptions only come into play once Claude has decided a tool is relevant, so a question it thinks it can answer from memory never reaches the knowledge base. That is the difference to look for.

In addition, I am thinking of the following:
Remove the `ask` and `analyze` and give the sandbox as a tool. Then the skill becomes important because it is what explains how everything is glued together. With it (I don't know if you have tried `analyze`) the llm will be able to traverse programmatically your entire db. All sorts of things are possible then which are not possible with just search including your metadata filtering and what not. 

With regards to visual grounding. Not yet I think, chunk ids do not leave the server, so there is nothing to hand `visualize_chunk`. But we can revisit and see how to address this.

PR #610, opened 2026-09-07
Keep footnotes and matched items in expanded context

### LINKED ISSUE #609: Context expansion drops the item the result matched on, when that item is a footnote
`_NOISE_LABELS` in `haiku/rag/context.py` excludes footnotes from expanded content whenever
the window contains a section header. `_build_expanded_result` applies that filter to every
item in the range, including the item the result matched on. A search that hits a footnote
therefore returns the surrounding section but with that footnote removed, so the passage never
contains the text that led to the result. Also, when a body paragraph is the hit, the same filter
strips the footnotes from the section returned with it.

I think there are two different questions:

1. Whether an item a result matched on should *ever* be filtered out of that result's passage.
   That looks wrong to me regardless of the label involved.
2. Whether `footnote` belongs in `_NOISE_LABELS` by default. Lawyers do love their footnotes, which are often carry substantial material (including legal authorities and cross-references), so would normally want to see them in a returned result.  If nothing else, having `_NOISE_LABELS_` be a configurable set would help.

If you would like a PR, let me know what direction you want to go in.

ggozad:
Closes #609.

- `_build_result` no longer drops the items a result matched on. Noise is a set of positions from `_noise_positions(items, matched_positions)`, computed per merged group and per split result. `_expand_outward` takes `noise: AbstractSet[int]` in place of `skip_noise`, `_build_result` takes it in place of `has_sections`. A matched item stays in the content and counts toward the budget.
- `footnote` removed from `_NOISE_LABELS`. The set is `page_header`, `page_footer`, `document_index`. No configuration option.
- Body footnotes are not anchored to the paragraph that cites them: docling links footnotes only to tables, pictures and code, and the PDF pipeline keeps no superscript marker.

Before merge: paired `orb_text` eval.


ggozad:
Waiting for evaluations to verify no regressions, will merge as soon as this is done unless there are surprises. cc @lawrenceakka

PR #617, opened 2026-09-08
Recover InlineGroup text for import_document/import_documents

### LINKED ISSUE #615: Markdown and HTML ingestion stores an empty item for every inline group
docling's markdown and HTML backends wrap a paragraph, list item or heading that
contains inline markup (code spans, links) in an `InlineGroup` whose children carry the
text. `DoclingDocument.iterate_items()` skips the group and yields the children, so
`extract_items` stores the parent with empty text and the children as separate items,
with no record that they belong together.

Measured with docling 2.124.0, `do_ocr=False`:

| file | InlineGroups | empty-text items | breakdown |
|---|---|---|---|
| `tests/data/doclaynet.pdf` | 0 | 0 of 498 | — |
| `CHANGELOG.md` | 1151 | 1151 of 7882 | 1021 `list_item`, 130 `section_header` |
| `README.md` | 43 | 34 of 184 | 34 `list_item` |
| 4-line HTML probe | 2 | 1 of 9 | 1 `list_item` |

Empty items equal inline groups exactly on `CHANGELOG.md`. PDFs are unaffected: the
layout pipeline has no inline-markup layer.

Three consequences:

1. **Empty items in the store.** `document_items` holds 1151 rows with no text for
   `CHANGELOG.md`, 130 of them `section_header`. `build_toc` and
   `get_document_outline` render those headings as blank, and `ChunkMetadata.headings`
   carries `""`, giving breadcrumbs like `Changelog >  > Added`.

2. **Split passages.** `context.py` `_build_result` joins items with `"\n\n"`, so a
   sentence containing inline code expands as three blocks: `By default,` /
   `haiku.rag` / `uses the configured embedder.`

3. **Fenced blocks mid-sentence in chunk text.**
   `HybridChunker._make_chunk_from_doc_items` re-serializes each doc item on its own
   when it merges or splits chunks, so a `code` child becomes a fenced block inside the
   chunk's own text. `HierarchicalChunker` renders the group inline correctly;
   `processing.chunker_type` defaults to `hybrid`.

Empty headings occur only with markdown reference-style links. The empty `list_item`
case affects markdown and HTML both, and is the larger share.

Pre-existing, and unchanged by the docling 2.102.2 → 2.124.0 move: 130 empty headings
before and after. Only newly ingested documents would change under any fix.

AmirF194:
Rescoped this to `import_document`/`import_documents`, per your comment: they store the caller's `DoclingDocument` and chunks as given, so a chunk's `doc_item_refs` index into it directly, and #621's flatten only runs at conversion time. A heading or list item docling pushed into a child `InlineGroup` still landed with an empty `document_items` row there. `extract_item_text` recovers it from that group, without touching any refs.

The group's own child rows still get stored too, so the recovered item and its fragments both land in expanded context. That tradeoff is now in the CHANGELOG entry.

Fixed the docstring on `_doc_with_inline_group_list_item`: it claimed the hand-built shape matches any paragraph, heading, or list item. A body-level paragraph's group hangs off `#/body` with no owner item, so that part was wrong.

Added `test_import_document_recovers_inline_group_text_without_moving_refs`, through `client.import_document` rather than `extract_item_text` directly. It builds a list item with an `InlineGroup` child by hand, imports it with a chunk whose `doc_item_refs` cites the list item, and checks the recovered text plus the untouched ref. Pulled the `InlineGroup` branch out locally first to confirm this test fails without it, then restored it: `tests/store/test_document_items.py`, 48 passed. `ruff check` and `ty check` clean on the touched files.

Dropped `Fixes #615` from the title and commit, it's `Refs #615` now, since #621 is what closes it.

One heads-up: #621's diff also lands in this same `### Fixed` block under `[Unreleased]`, cutting a `0.83.0` release there, so whichever of these merges second will need a routine rebase on `CHANGELOG.md`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@
 - Conversion flattens each inline group of text runs into the item that owns
   it. Hyperlinks are dropped from heading text. A full `haiku-rag rebuild`
   applies the fix to existing databases.
+- `import_document`/`import_documents` store an item whose text docling
+  placed in a child `InlineGroup` with that group's text. The group's own
+  rows remain. An inline image inside the group is dropped instead of
+  stored as docling's markdown placeholder. Table rows in `document_items`
+  are now stored without markdown escaping too.
 
 ## [0.84.0] - 2026-09-10
 
@@ -143,9 +148,6 @@
   not checked against the deadline.
 - Context expansion keeps the item a result matched on when it carries a
   noise label, and counts it toward the character budget.
-- `import_document`/`import_documents` store an item whose text docling
-  placed in a child `InlineGroup` with that group's text. The group's own
-  rows remain.
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -143,12 +143,9 @@
   not checked against the deadline.
 - Context expansion keeps the item a result matched on when it carries a
   noise label, and counts it toward the character budget.
-- `import_document`/`import_documents` store the caller's `DoclingDocument`
-  as given, so a heading or list item docling had pushed into a child
-  `InlineGroup` still landed with an empty `document_items` row there.
-  `extract_item_text` now recovers it from that group. The group's own
-  child rows stay in the table too, so the recovered item and its fragments
-  both land in expanded context.
+- `import_document`/`import_documents` store an item whose text docling
+  placed in a child `InlineGroup` with that group's text. The group's own
+  rows remain.
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -143,6 +143,12 @@
   not checked against the deadline.
 - Context expansion keeps the item a result matched on when it carries a
   noise label, and counts it toward the character budget.
+- `import_document`/`import_documents` store the caller's `DoclingDocument`
+  as given, so a heading or list item docling had pushed into a child
+  `InlineGroup` still landed with an empty `document_items` row there.
+  `extract_item_text` now recovers it from that group. The group's own
+  child rows stay in the table too, so the recovered item and its fragments
+  both land in expanded context.
 
 ### Removed
 
--- a/haiku_rag_slim/haiku/rag/store/models/document_item.py
+++ b/haiku_rag_slim/haiku/rag/store/models/document_item.py
@@ -82,11 +82,28 @@ def extract_item_text(
       so pictures carry meaningful prose into chunk text and survive
       ``expand_with_items``' ``if item.text:`` filter; otherwise fall back to
       the picture's caption text.
+    - Items whose own text is empty because docling pushed mixed inline
+      content (e.g. a paragraph or list item containing a code span or link)
+      into a child InlineGroup instead: serialize that group, the same way
+      docling itself renders it back to markdown.
     """
-    from docling_core.types.doc.document import PictureItem, TableItem
+    from docling_core.types.doc.document import InlineGroup, PictureItem, TableItem
 
     if text := getattr(item, "text", None):
         return text
+
+    def _serialize(target: Any) -> str | None:
+        try:
+            serializer = get_serializer() if get_serializer is not None else None
+            if serializer is None:
+                from docling_core.transforms.serializer.markdown import (
+                    MarkdownDocSerializer,
+                )
+
+                serializer = MarkdownDocSerializer(doc=docling_doc)
+            return serializer.serialize(item=target).text
+        except Exception:
+            return None
 
     if isinstance(item, PictureItem):
         if description := _picture_description_text(item):
@@ -94,18 +111,12 @@ def extract_item_text(
         return _picture_caption_text(item, docling_doc)
 
     if isinstance(item, TableItem):
-        try:
-            if get_serializer is None:
-                from docling_core.transforms.serializer.markdown import (
-                    MarkdownDocSerializer,
-                )
+        return _serialize(item)
 
-                serializer = MarkdownDocSerializer(doc=docling_doc)
-            else:
-                serializer = get_serializer()
-            return serializer.serialize(item=item).text
-        except Exception:
-            pass
+    if item.children:
+        first_child = item.children[0].resolve(docling_doc)
+        if isinstance(first_child, InlineGroup):
+            return _serialize(first_child)
 
     return None
 
--- a/haiku_rag_slim/haiku/rag/store/models/document_item.py
+++ b/haiku_rag_slim/haiku/rag/store/models/document_item.py
@@ -98,9 +98,13 @@ def extract_item_text(
             if serializer is None:
                 from docling_core.transforms.serializer.markdown import (
                     MarkdownDocSerializer,
+                    MarkdownParams,
                 )
 
-                serializer = MarkdownDocSerializer(doc=docling_doc)
+                serializer = MarkdownDocSerializer(
+                    doc=docling_doc,
+                    params=MarkdownParams(escape_underscores=False, escape_html=False),
+                )
             return serializer.serialize(item=target).text
         except Exception:
             return None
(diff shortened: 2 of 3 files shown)


ggozad:
Thanks for this your diagnosis is spot on. 

I started working on this yesterday but hadn't pushed anything, not expecting a PR on it. I am sorry about that, I would still welcome you as a contributor. I went with fixing it at the conversion boundary instead (#619). The catch with `extract_item_text` is that the chunkers read the `DoclingDocument`, not `document_items`, so chunk text and breadcrumbs stay broken either way. I ran your branch against a small probe and the chunk still came out with a fenced code block mid-sentence and headings `['Sample', '', 'Added']`. A plain paragraph also has no owner item to repair, since its group hangs off `#/body`, so it stays three rows. And the text you recover sits next to the fragment rows, which are still stored and still get assembled by context expansion.

Easy to miss: building the document by hand with `add_group(label=GroupLabel.INLINE)` always puts an item in the parent slot. A real markdown conversion doesn't.

One place your approach is the right one is `import_document`. We never convert that document, and we can't flatten it either as the caller's chunks carry `doc_item_refs` into it as passed, and flattening renumbers them. Deriving the text without touching refs is exactly what fits. If you want to rescope #617 to that, I'll take it.

Not that #619 is based on #614 which I am hoping to release today.

AmirF194:
Read #619, it doesn't touch extract_item_text, so nothing here conflicts with it. Checked import_document myself: it stores the caller's docling_document as given because the chunks' doc_item_refs index into it, so flattening isn't an option on that path and this fix is the right one for it specifically. #617's body only ever claimed the document_items fix, not chunk correctness, so I don't think the diff needs to change. Fine to leave it scoped as is, or I can retitle it to call out import_document explicitly if that reads clearer on your end.

ggozad:
Happy to take this scoped to `import_document`:

- Drop `Fixes #615`, it's in the commit message as well as the body. `Refs #615` is fine; #619 closes it.
- Retitle and reword the description to what the change actually does: recover an item's text from its child `InlineGroup` for documents that never went through conversion. The current body ties it to breadcrumbs and mid-sentence chunk splits, which it doesn't reach.
- Same for the CHANGELOG entry. After #619 a converted document has no inline groups left, so its intended remaining path is externally supplied documents through `import_document` or `import_documents`.
- An integration test through `import_document` would be great. Be aware that once #619 is in, `rag.convert()` won't hand you an unflattened document any more, so you'll need to call docling's `DocumentConverter` directly or patch `haiku.rag.converters.docling_local.flatten_inline_groups` off while building the fixture. `tests/test_rebuild.py::test_rebuild_rechunk_normalizes_a_stored_document` in #619 does the latter if you want a template. While you're there, the docstring saying the hand-built list-item shape is what "any paragraph, heading or list item" produces isn't right, a body-level paragraph's group hangs off `#/body` with no owner item at all.

One caveat worth documenting: the fragment rows stay, so on that path the recovered text and its fragments both end up in expanded context. I'm willing to accept that tradeoff, but it should be mentioned somewhere.

ggozad:
FYI #619 was closed when the base branch was deleted yesterday so it's replaced by #621

AmirF194:
Retitled and rewrote the body to scope this to `import_document`/`import_documents`, dropped `Fixes #615` for `Refs #615`, and reworded the CHANGELOG entry the same way. Left the fix itself as-is since it's a general helper, not import-specific code.

Added `test_import_document_recovers_inline_group_text_without_moving_refs`, going through `client.import_document`: builds a list item with an `InlineGroup` child, imports it with a chunk citing that item's ref, checks the recovered text and that the ref didn't move. Confirmed it fails without the `InlineGroup` branch and passes with it restored.

Fixed the docstring you flagged: a body-level paragraph's group has no owner item, so it's not the same shape as the hand-built list item.

The fragment-rows tradeoff is now a CHANGELOG line, since that's where the fix itself is described.

ggozad:
Thanks, this is now correct and I'll take it scoped like this.

One update from my side that makes your description more accurate than when you wrote it: #621 has been narrowed so that `flatten_inline_groups` runs only at conversion. `rebuild --rechunk` no longer normalizes stored documents, and `import_document` / `import_documents` stay untouched. So after both land, a converted document has no inline groups left except the ones #621 deliberately leaves alone (a group carrying an inline picture, and a couple of shapes no backend produces), and everything else with an inline group under an empty owner arrives through import. That is exactly the path your change covers, and it also means your `extract_item_text` branch will see those inline-picture groups from our own converters, which is fine.

Three things before I merge your which has to follow #621 anyway:

1. The serializer defaults escape. `MarkdownDocSerializer(doc=docling_doc)` renders with `escape_underscores=True` and `escape_html=True`, so a recovered row reads `snake\_case` and `&amp;` where the runs held `snake_case` and `&`. We hit the same trap in #621. Pass `MarkdownParams(escape_underscores=False, escape_html=False)` when you construct the fallback serializer, and add an underscore and an ampersand to the plain run in your fixture so the test pins it.

2. `item.children[0]` assumes the inline group is the first child. A list item can carry a nested list group beside its inline group, and your own fixture builds that shape with `second_child_group=True`. Look for the first `InlineGroup` among the children rather than checking position zero.

3. CHANGELOG. Keep the entry to what changed, one or two sentences, no "so": something like "`import_document` and `import_documents` store an item whose text docling placed in a child `InlineGroup` with that group's text. The group's own rows remain." #621 is landing first, so you will need the rebase onto its `### Fixed` block that you already flagged.

The `_serialize` helper swallowing every exception is the existing table behaviour, so I won't ask you to change it here.

In all honesty, #621 has been frustrating and difficult to deal with in particular when it comes to supporting importing documents outside haiku.rag's control. Is this a feature you use? It was introduced a long while ago at a time when ingestion was a lot simpler and I am considering removing it altogether in the future.

AmirF194:
All three addressed at ba40d947.

1. Both serializers (the fallback in `_serialize` and the shared one in `extract_items`) now build with `MarkdownParams(escape_underscores=False, escape_html=False)`. Added `snake_case & ` to the fixture's plain run and a test pinning the unescaped output.

2. Swapped `item.children[0]` for a loop over all children, returning the first `InlineGroup`. Added a test with the nested `ListGroup` placed before the `InlineGroup` to pin it (it returned `None` on the old code).

3. Changelog entry trimmed to your wording.

Rebase onto #621 once it lands, that's still open.

ggozad:
Thanks!

I merged #621, so please rebase onto `main`. The CHANGELOG hunk will conflict; your entry goes under the current `[Unreleased]`, `0.83.0` and `0.84.0` stay as they are.

Two small things while you're in there, then I'll merge:

1. After #621, the one shape from our own converters that still reaches your branch is a bullet or heading carrying an inline image, which the flatten deliberately leaves as a group. Your recovery then stores `Click <!-- image --> icon to start.` as the list item's text. Add `image_placeholder=""` to both `MarkdownParams` so it stores `Click icon to start.` instead. The placeholder is docling's markdown rendering detail, not text.

2. Changing the shared serializer in `extract_items` means table rows in `document_items` are stored without markdown escaping too. That's the right behaviour and consistent with #621, but it changes stored text for every document with tables on the next rebuild, so the CHANGELOG should say it. One sentence on your entry: "Table rows in `document_items` are stored without markdown escaping."

3. The comment in `test_import_document_recovers_inline_group_text_without_moving_refs` that starts "The tradeoff ggozad asked to have written down" should state the behaviour rather than the conversation, e.g. "The group's children are stored as their own rows beside the recovered item."

Also you didn't tell me if you are actually using import_document(s)

AmirF194:
Rebased onto main after #621. The CHANGELOG hunk didn't actually conflict, but the entry had landed under 0.83.0 instead of Unreleased, so I moved it and folded in both notes below it.

1. Added `image_placeholder=""` to both `MarkdownParams` calls, confirmed it drops the placeholder rather than leaving alt text behind.
2. Table rows losing markdown escaping is now in the changelog entry, one sentence as you suggested.
3. Reworded the test comment to describe the behavior instead of the conversation.

50 of the 51 tests in the file pass in a clean container; the one failure needs `opencv`, which this verify command doesn't install, and it's unrelated to this diff (pre-existing test on main). ruff check and ruff format both clean.

I found this from your issue, not from using `import_document` myself, so I can't speak to real usage patterns beyond what the issue describes.

ggozad:
Thanks again Amir!
Will merge as soon as CI clears. 
To have all the extras needed for testing installed, you need to `uv sync` in your repo root.

AmirF194:
Thanks for the back and forth on this, the rescope and the escaping/placeholder details are both better for your review.

PR #635, opened 2026-09-11
Support email file formats

ggozad:
Support for .eml and .msg extensions already supported by docling.


docs/configuration/processing.md
```diff
@@ -268,7 +268,7 @@ Per-image failures (404, timeout, oversized, unreadable) leave that picture as a
 | `.html` / `.xhtml` | n/a (markup-based) | n/a | ✅ on embedded pictures | ✅ |
 | `.md` / `.qmd` / `.rmd` | n/a | n/a | ✅ on embedded pictures | ✅ (only `<img>` HTML blocks; native `![alt](url)` syntax is not fetched by docling) |
 | `.docx` / `.pptx` | n/a | n/a | ✅ on embedded pictures | n/a |
-| Other (`.csv`, `.xlsx`, `.adoc`, `.tex`, `.xml`) | n/a | n/a | n/a | n/a |
+| Other (`.csv`, `.xlsx`, `.adoc`, `.tex`, `.xml`, `.eml`, `.msg`) | n/a | n/a | n/a | n/a |
```
tseaver, line 271: Do we want to remove `.csv` and `.xlsx` here?

haiku_rag_slim/haiku/rag/converters/docling_local.py
tseaver, line 110: Here too?

ggozad: I wouldn't, just because I can see how some people might have a use-case around this.
The list was hand-maintained because at the time there was alternative backends than docling and then it remained hand-maintained. I guess it should come from docling. There's another that translates code to markdown  which might also be replaceable.

PR #637, opened 2026-09-14
Add `sources` param to `list_documents` tool

### LINKED ISSUE #634: Sources for list_documents?
The `list_documents` mcp tool does not have a `sources` parameter, unlike all the other mcp tools. It is useful to be able to list docs by source (at least for me).  Passing `source=None` could return all docs as at present.

lawrenceakka:
Closes #634


ggozad (changes requested):
Thank you again @lawrenceakka !!!
Some issues to resolve:

* `sources` is ignored when the client covers one database

`HaikuRAG.list_documents` threads `sources` through the `covers_multiple` branch only. As a result, on the default configuration of one database the new parameter does nothing.

For example for one database holding two documents:

```
list_documents()                   -> 2
list_documents(sources=[])         -> 2   expected 0
list_documents(sources=["nope"])   -> 2   expected UnknownDatabaseError
search_documents(sources=["nope"]) -> UnknownDatabaseError
```

`HaikuRAG.search` has the guard already, before its single-database path:

```python
if not await self.clients_covering(sources):
    return []
```

The same line before `list_documents`' single-session return would cover both cases. `clients_covering` raises on a name this client does not cover and returns nothing for an empty selection.

* Typo in the tool docstring

The Args block documents `source`, singular, described as one collection. The parameter is `sources` and takes a list. That docstring is the tool description the model reads, and Claude Desktop shows nothing else, so a model will call `source=` and get an unexpected-argument error.

Delete both lines. `search_documents` and `execute_code` document neither `filter` nor `sources` in their Args. Both descriptions come from the `Annotated` `Field` on the shared `Filter` and `Sources` types, and `Sources` already reads "Collections to use, by name. All of them by default."

* Tests 
The change is to the public client API, but nothing was added to the multi-database document tests, where the listing, paging and filter tests live, and no test covers a single database. That gap is what hid the first issue. The new test also asserts the empty selection alongside the restriction, which its docstring does not name.

* Docs

 The storage configuration page still says `search`, `ask` and `analyze` take `sources`. `list_documents` belongs in that sentence.

ggozad:
Thanks again @lawrenceakka!

PR #645, opened 2026-09-21
Move UnsupportedSourceError out of the client package

### LINKED ISSUE #644: ingester run-batch --dry-run takes an unnecessarily long time.
... on my (slow) machine, nearly 9 s, even if there are no sources.

The cause is the import chain. `sources/fs.py`, `sources/s3.py`, `sources/registry.py`, `converters/pdf_split.py` and `client/documents.py/processing.py` all raise `UnsupportedSourceError`, which is currently imported from `haiku.rag.client.exceptions`. That is a submodule of the `client` package (obviously), so `client/__init__.py` is always run first.  That in turn imports `Store` and `get_embedder`. `get_embedder` pulls in pydantic_ai's agent and MCP stack, the openai SDK, and `lance/pyarrow`.  None of that is relevant to a `--dry-run`, and costs about 7 seconds.

The solution would be to move `UnsupportedSourceError` out of the `client` package.  In fact, I think it does not really belong there anyway. It is raised by three unrelated subsystems (sources, converters, client) and caught centrally by the ingester's job classifier. So it looks to me to be similar to `circuit_breaker.py`, which already lives at the top level (`haiku/rag/circuit_breaker.py`). Moving `UnsupportedSourceError` to a comparable top-level module (or something like `haiku.rag.exceptions`), with `client/exceptions.py` re-exporting it for backward compatibility if necessary, fixes the import cost for every current and future caller that doesn't need the rest of client.

AmirF194:
`UnsupportedSourceError` lived in `haiku.rag.client.exceptions`, so importing it from a source or converter forced `haiku.rag.client.__init__` to run, pulling in lancedb, pyarrow and pydantic_ai. That's most of what `--dry-run` was paying for: importing `haiku.rag.sources.registry` alone took 5.5s versus 0.02s for a plain `import json`.

Moved the class to a new top-level `haiku.rag.exceptions`. `client.exceptions.UnsupportedSourceError` now just re-exports the same object, so any existing `except UnsupportedSourceError` still catches it (confirmed `Old is New`). Repointed the four source/converter call sites plus `ingester/workers/pipeline.py`, which imports it too but wasn't in the list above.

Added a subprocess-based test next to `test_cli.py`'s existing `test_importing_cli_does_not_load_heavy_dependencies`, same shape: import `haiku.rag.sources.registry` in a fresh process and assert lancedb/pyarrow/pydantic_ai never load. The new test doesn't pass without the change. Ran the rest of the sources/cli/converters/ingester suite too, no regressions, and ruff/ty are clean on the changed files.

Used AI for the implementation; read through the diff and the test before opening this.

Fixes #644


ggozad:
Thanks for picking this up, but this does not make the dry run faster: 2.4s before, 2.5s after on my machine for the reasons I described in the issue

I will do that in a separate PR. Closing this one.

PR #648, opened 2026-09-21
Split the ingester's store phase into embed / extract / write spans

bd-mkt:
# Split the ingester's store phase into embed / extract / write spans

## Why

`document.store` was a sum, not a phase. It wrapped `ensure_chunks_embedded`,
the `extract_items` thread hop, *and* the LanceDB write, so "storing took 40s"
could equally mean a slow embedder, a picture-heavy item extraction, or write
contention. The only child span was `document.embed`, and `extract_items` had
none at all, so subtracting the embed child to recover write time gave the
wrong answer on exactly the documents where it mattered.

Three further gaps fell out of the same review:

- The span sat at the ingest call site, so `create_document`,
  `import_document(s)` and `update_document` got no store span — their
  embedding showed up with no write beside it.
- `document.embed` was skipped entirely when every chunk arrived
  pre-embedded, leaving a hole in the trace on the import path.
- `client/processing.py` was the one module in the package doing a bare
  `import logfire` instead of importing the scoped instance from
  `haiku.rag.telemetry`, so `document.embed` shipped under
  `scope.name = 'logfire'`. A `scope.name = 'haiku.rag'` filter — which the
  `debug-ingestion` skill documents as catching the whole tree — silently
  dropped it.

## What changed

The pipeline phases are now siblings, each timing one thing:

```
ingester.job
├─ document.fetch      bytes, content_hash
├─ document.convert
├─ document.chunk      chunks_created
├─ document.embed      chunks, chunks_embedded, images, batch_size, batches
├─ document.items      uri, pictures, items
└─ document.store      op, document_id, uri, chunks, items
   └─ store.write_txn  lock_wait_ms
```

Their durations add up to the job instead of nesting, so a slow
`document.store` now means a slow write.

### `client/processing.py`

- Import the scoped `logfire` from `haiku.rag.telemetry`.
- `document.embed` is emitted unconditionally, including when nothing needs
  embedding (`chunks_embedded=0`), so the phase is never missing from a trace.
- New attributes: `chunks` (total on the document), `chunks_embedded`,
  `images`, `batch_size`, and `batches` — the number of provider round trips
  `embed_chunks` will make, since text goes out in `batch_size` groups and
  images one at a time. Slow with few `batches` is provider latency; slow with
  many is a batch-size problem. Attributes rather than per-batch child spans:
  a 2000-chunk PDF would otherwise flood the trace.

### `client/documents.py`

- The `document.store` span moved from the two call sites in
  `_ingest_fetch_result` into `_store_document_with_chunks`,
  `_update_document_with_chunks` and `_store_documents_with_chunks`, and now
  wraps only the `write_transaction` block. Every write path is traced the
  same way as an ingest.
- New `_extract_items_observed` helper wraps the `extract_items` thread hop in
  a `document.items` span (`pictures`, `items`, `uri`).
- On the create path, `op` is set *after* the transaction: a concurrent ingest
  of the same URI is only resolved under the write lock, so `op` is `update`
  whenever that re-check found an existing row.
- On an update carrying no docling document the `items` attribute is absent
  rather than `0` — the existing items are preserved, not replaced with none.
- The batch path reports `op=create_batch` with `documents=N`.

### `store/engine.py`

- `write_transaction` opens a `store.write_txn` span *after* acquiring
  `_write_lock`, carrying the time it spent queued as `lock_wait_ms`. The
  caller's `document.store` therefore measures lock wait plus write, and this
  one isolates the write. With concurrent ingester workers on one lock, that
  split is usually the answer to "why is storing slow".
- Note this fires for **every** writer, including vacuum and `create_tag`, so
  `store.write_txn` also appears outside the document pipeline.

## Breaking change for dashboards

`document.store.duration` changes meaning: it drops from "embed + extract +
write" to "write". Any saved Logfire view, alert threshold or report built on
that span needs re-baselining after deploy. Nothing in the API or on-disk
format changes.

## Tests

New `tests/test_store_telemetry.py`, 11 cases, driven through a `TestExporter`
against a stub embedder (no cassette, no network):

- the three phases share a parent and none contains another;
- `store.write_txn` nests under `document.store` and carries `lock_wait_ms`;
- `document.embed` counts what it embedded, and is still emitted with
  `chunks_embedded=0` when every chunk arrived pre-embedded;
- `document.store` reports `op=create`, `op=update` on the update path, and
  `op=update` when the under-lock re-check finds the URI already taken;
- an update with no docling document emits no `document.items` and no `items`
  attribute;
- the batch path reports `op=create_batch` with `documents=2`;
- all three spans carry `scope.name = 'haiku.rag'`.

No existing test asserted on these span names, so nothing was rewritten to
suit the change.

## Docs

- `.claude/skills/debug-ingestion/SKILL.md` — rewritten pipeline vocabulary,
  plus three new canned queries: where a job's time goes, whether a slow store
  is contention or throughput (joining `store.write_txn` to its parent for the
  URI), and whether slow embedding is provider latency or batch count.
- `docs/ingester.md` — the span tree section.
- `CHANGELOG.md` — including the re-baselining warning above.

## Verification

- `uv run ruff check` / `ruff format --check` — clean.
- `uv run ty check` on the changed files — clean. Repo-wide `ty` reports three
  pre-existing errors (`time.tzset`, `signal.SIGKILL`) in files this PR does
  not touch; they are POSIX-only stdlib members flagged when checking on
  Windows.
- `uv run pytest tests/test_store_telemetry.py` — 11 passed.
- Full suite (`-m "not integration"`) on Windows, compared against a baseline
  run of the same suite in a clean worktree at HEAD:

  | | passed | failed | errors |
  |---|---|---|---|
  | baseline (HEAD) | 2771 | 21 | 30 |
  | this branch | 2782 | 21 | 30 |

  The delta is exactly +11 passes — the new tests — and the failure sets are
  identical test-for-test (`diff` of the two `-rf` lists is empty). The 21
  pre-existing failures and 30 errors are Windows-host artifacts in areas this
  PR does not touch: `UnicodeDecodeError: 'charmap'` at fixture setup (host
  locale encoding), docling-serve tests needing a live service, and
  symlink / sandbox worker-crash tests. Linux CI remains the authoritative
  check.


ggozad (changes requested):
Thanks. Splitting embedding out of `document.store` is the right change, and the `document.embed` scope fix is real (it was the last bare `import logfire` in the package). Requesting changes, mostly to the docs, plus two code changes.

## 1. Replace the `store.write_txn` span with a `lock_wait_ms` attribute

`lock_wait_ms` is the useful measurement: it is the only place a trace shows time spent queued behind vacuum, tag operations and metadata-only updates, which take `_write_lock` directly and emit no span. The span around it adds little:

- `write_transaction` has four callers. Three already wrap it in `document.store`, and the fourth is `delete_document`.
- It carries no `uri`, which is why the skill needs a self-join on `records` to find which document waited.
- "`document.store` minus `store.write_txn` is contention" does not hold. On a local ingest I measured `document.store` 24 ms, `store.write_txn` 15 ms, `lock_wait_ms` 0.004.

Setting the attribute on the current span keeps the measurement, drops a span per write and leaves the rollback block un-indented:

```python
from opentelemetry import trace

waiting_since = monotonic()
async with self._write_lock:
    trace.get_current_span().set_attribute(
        "lock_wait_ms", (monotonic() - waiting_since) * 1000
    )
    versions = await self.current_table_versions()
    ...
```

It lands on `document.store` next to `uri` and `op`, and on `ingester.job` for a DELETE job. With no recording span it is a no-op. One limit to state in the docs: `_write_lock` is in-process, so contention between two ingester processes on one database is not measured.

## 2. `document.store` and `document.items` changed which `uri` they report

The spans moved into `_store_document_with_chunks` / `_update_document_with_chunks`, so they now report `document.uri` where `document.store` used to report `result.uri`. The two differ: on macOS, for a file under `/var/folders/...`, `document.convert` and `document.chunk` report `file:///private/var/.../x.md` while `document.store` and `document.items` report `file:///var/.../x.md`. Anything grouping or joining spans by `uri` splits one document into two keys. Either pass the fetch URI down or leave `uri` off these two spans. `ingester.job` already carries it.

## 3. Docs describe behaviour that does not exist

`docs/ingester.md`, `SKILL.md` and the PR body say every writer emits `store.write_txn`, including vacuum and tagging. They don't: `vacuum` (`engine.py:393`), tag create/delete/restore (`engine.py:831/923/985`) and the three metadata writers in `documents.py` hold `_write_lock` directly. With item 1 these paragraphs become: "`document.store` carries `lock_wait_ms`, the time spent waiting on the in-process write lock."

## 4. The skill's "this sums" query double-counts

`span_name LIKE 'document.%'` matches `document.convert_slice`, which nests inside `document.convert` (`pdf_split.py:112`), so with `processing.split_pages` on, conversion time is counted twice. Use an explicit `IN (...)` list. "Durations add up to the job" is also not quite true: `get_all_picture_data` on the update path and `_prepare_and_title` on the create path sit in no span.

## 5. Drop `batches` from `document.embed`

`ceil(texts / batch_size) + images` copies `embed_chunks`' batching logic into `ensure_chunks_embedded`. It is correct today, but it will report the wrong number when that logic changes, and no test ties the two together. `chunks_embedded`, `images` and `batch_size` are already on the span, so the count can be computed in SQL.

## 6. CHANGELOG

`StoreEngine.write_transaction` does not exist; the class is `Store`. The project's CHANGELOG entries are one-line fragments without rationale or narration of the old behaviour ("It used to span...", "the one span in the package that was"). With the changes above, roughly:

```
### Added
- `document.items` span.
- `document.embed` reports `chunks_embedded`, `images` and `batch_size`, and is emitted when no chunk needs embedding.
- `document.store` reports `chunks`, `items` and `lock_wait_ms`; the batch path reports `op=create_batch` and `documents`.

### Changed
- `document.store` covers the LanceDB write only; dashboards and alerts on `document.store` duration need re-baselining.

### Fixed
- `document.embed` carries `scope.name = haiku.rag`.
```

## 7. Comments

AGENTS.md keeps rationale out of code comments. The `_extract_items_observed` docstring ("Kept separate from the write so...") and the comments starting "The span is emitted even when...", "`op` is only knowable after...", and "Absent rather than zero..." should each shrink to the constraint, or go.

## 8. Tests: two missing

- **The ingest path.** Every test calls `import_document`, `import_documents` or `update_document` with `mem://` URIs, so `_ingest_fetch_result`, the path the ingester uses, is never exercised. That is why item 2 passes. Add a test that ingests a file from `tmp_path` through `create_document_from_source` and asserts `document.store`'s `uri` equals `document.convert`'s.
- **Contention.** `assert lock_wait_ms >= 0` cannot fail. Hold the write lock, start the write, release, and assert `lock_wait_ms` on `document.store` is greater than zero. Use `ObservedLock` and `assert_waiting_for_lock` from `tests/locks.py` (see the rebase note below): swap it in with `monkeypatch.setattr(store, "_write_lock", lock)`, hold it, and confirm the write is actually waiting before releasing. A `sleep` followed by `assert not task.done()` cannot tell a held lock from a task that has not run yet.

## Rebase

#650 reorganised the test suite on main, so this needs a rebase:

- `CHANGELOG.md` conflicts: main's `[Unreleased]` now has a `### Changed` section. Add your `Changed` entries to it, not as a second `### Changed` heading.
- Move the file to `tests/client/test_telemetry_spans.py`. Tests now live by source subsystem, these drive `HaikuRAG`, and `tests/test_telemetry.py` already tests `telemetry.configure()` at a different layer.
- `tests/locks.py` is on main now, for the contention test above.
- `pytest.mark.asyncio` is no longer used (auto mode). Your file doesn't use it; keep it that way for the new tests.
- CI's coverage-gated lane is `-m "not slow and not integration"`, so leave these tests unmarked: they cover new source lines and need no network.

Nothing else conflicts. I rebased this branch onto main locally (see below), moved the file, and ran the fast lane.

## Verified on macOS

- `ruff check`, `ruff format --check`, repo-wide `ty check`: clean.
- The branch as submitted, `pytest -m "not integration" --cov=haiku`: 2996 passed, 1 skipped, coverage 100%.
- The fixture's global `logfire.configure()` does not break later tests in the same worker: `tests/test_store_telemetry.py` followed by `tests/test_mcp.py` in one process, 84 passed.
- Rebased onto `a493b8f67ed5` with the file moved as above (the CHANGELOG conflict left unresolved; it does not affect the run): the file alone, 11 passed; the fast lane (`-m "not slow and not integration" --cov`), 2978 passed, 1 skipped, coverage 100%.
- A real ingest (create, then update) on main and on this branch, spans exported and compared. Items 1 and 2 come from those traces.

ggozad (approved):

